PR #17, opened 2023-12-04
mark get_field_names() [[gnu::no_sanitize_undefined]]

### LINKED ISSUE #16: Link error with -fsanitize=undefined
The following program fails to link on g++ 13.2.0 x86-64 Linux (Ubuntu):
```c++
#include <rfl/internal/get_field_names.hpp>
struct A { int i; };
int main() { rfl::internal::get_field_names<A>(); }
```
error: 
```
/tmp/ccKMlHON.o:a.cpp:function auto get_field_names<A>():(.text._Z15get_field_namesI1AEDav+0x5): error: undefined reference to 'fake_object<A>'
collect2: error: ld returned 1 exit status
```
might be related to #15 
reduced: https://godbolt.org/z/1jecWY947

possible workaround is to use `[[gnu::no_sanitize_undefined]]`

ecatmur:
fixes #16

### DIFF
--- a/include/rfl/internal/get_field_names.hpp
+++ b/include/rfl/internal/get_field_names.hpp
@@ -114,6 +114,9 @@ with open("generated_code5.cpp", "w", encoding="utf-8") as codefile:
 */
 
 template <class T>
+#if __GNUC__
+[[gnu::no_sanitize_undefined]]
+#endif
 auto get_field_names() {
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
I have implemented the proposed fix, so I don't think the PR is necessary any longer.

I haven't used your PR, because it would have likely created a merge conflict. Sorry about that, but thank you for your contribution. It is very appreciated!

PR #19, opened 2023-12-06
Reduce boilerplate

ChemiCalChems:
After looking around the codebase, I saw some boilerplate that could be easily removed.

There is more I think could also be removed but I'm unsure whether it would be acceptable or not so I would rather discuss it with someone in the project first. Get in contact with me if you're interested. 

On my machine the tests now compile a couple of seconds quicker and the binary size is reduced by 160K.


liuzicheng1987:
I have taken a look at your commits. I think say they make a lot of sense. Thank you very much. I will test this and then merge very soon.

liuzicheng1987:
@ChemiCalChems Sure, I am always open for suggestions. You can get in touch with me on Reddit or LinkedIn, whichever you prefer.

liuzicheng1987:
@ChemiCalChems , the tests run through on clang, gcc and MSVC. I am merging this. Thank you very much.

PR #31, opened 2023-12-16
Serialization of std::reference_wrapper

ecatmur:
Only for serialization - similarly to pointers, this should not be used for deserialization.

As discussed https://github.com/getml/reflect-cpp/pull/20#issuecomment-1857491501 this requires changing the implementation of bind_to_tuple, since std::ref() collapses on a std::reference_wrapper.

Checked tests run under g++ 11.4.0
Formatted additions with clang-format 16.0.0

### DIFF
--- a/include/rfl/internal/bind_to_tuple.hpp
+++ b/include/rfl/internal/bind_to_tuple.hpp
@@ -28,7 +28,7 @@ constexpr auto bind_to_tuple(T& _t, F&& f) {
 /*The following boilerplate code was generated using a Python script:
 def make_field_template(num: int) -> str:
     fields = ", ".join([f"f{i+1}" for i in range(num)])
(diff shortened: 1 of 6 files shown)


liuzicheng1987:
This looks like it makes a lot of sense. I‘m going to test it on clang, gcc and MSVC and if that checks out, I‘ll merge.

ChemiCalChems:
Thank for the contribution!

liuzicheng1987:
By the way, @zerolfx is in the process of building up a Github Actions pipeline...I really look forward to that. It is going to make testing and approving PRs a lot easier.

liuzicheng1987:
Tests compile and run on gcc, clang and MSVC. And you have even added a test for std::ref. Great contribution. Thank you so much.

Issue #37, opened 2023-12-21
Add compile-time regex support using Hanickadot's CTRE


ChemiCalChems:
It would be nice to add support for compile-time regular expressions for the following reasons as it would
- Allow for compile-time checking of validation regexes (pretty much impossible to push invalid regexes if the code doesn't compile)
- Quicker runtime validation of fields (since the regex is processed at compile-time)
- Allow for compile-time validation whenever we (eventually) support compile-time parsing of JSON

This last point seems like science-fiction, or like something that wouldn't ever be useful, but with the addition of `#embed` to C23 and its likely addition to C++ at some point (hopefully by C++26) it wouldn't be crazy to expect users to want to configure programs using inline / embedded JSON or other formats. There are already a few options for compile-time JSON parsing, so it's not unheard of, in which case, validation of JSON fields would be necessary too.

For all this, I propose (and will be working on a PR for) the addition of [Hanickadot's CTRE](https://github.com/hanickadot/compile-time-regular-expressions/) as a dependency, as well as tweaks in the validator code to get rid of the extremely slow `std::regex`. If you want to see the speed differences, please check [P1433R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1433r0.pdf), which received a very positive reaction from the C++ Standard's committee (hopefully we'll have a `constexpr` capable option for regex in the standard library soon).

liuzicheng1987:
@ChemiCalChems Looking forward to it!

liuzicheng1987:
@ChemiCalChems when you add the library as a dependency, please just copy over the header and mention that in the README, just like YYJSON.

This is part of the core functionality and should work out-of-the-box.

ChemiCalChems:
@liuzicheng1987 I was thinking of adding the library in as a `git submodule` for better trackability in versioning and updating in the future. It would also help (in the case adding YYJSON as a submodule too) our repo not be listed as mostly C just because of dependencies. Github also has tracking for submodules, listing them as a link to the repo they originally mirror, which is nice. The only problem is adding a whole submodule for a single header file. I've read up on `git sparse-checkout` but that seems to be a local-only solution, which sucks...

Another idea I had was having whatever submodules we want in a `submodule` directory and then symlinking whatever directories we do need wherever they happen to be needed. Maybe we could discuss this at some point.

liuzicheng1987:
@ChemiCalChems , thank you for your PR. I have merged it and we can now close this issue.

PR #42, opened 2023-12-28
Compile time regex (related to issue #37)

ChemiCalChems:
As discussed in #37, it would be beneficial to move over to compile-time regex. Hanickadot's CTRE allows for this. Here is the (hopefully) much awaited PR.

On my machine, tests now compile quicker and into smaller binaries. They also run marginally quicker, though I didn't test this with a big enough sample size to conclude anything other than anecdotally quicker runs.

```
~/src/reflect-cpp/build $ make clean > /dev/null && time make -j4 > /dev/null && du -b tests/json/reflect-cpp-json-tests 

real	1m30.108s
user	5m13.096s
sys	0m19.556s
31075952	tests/json/reflect-cpp-json-tests
~/src/reflect-cpp/build $ git switch main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
~/src/reflect-cpp/build $ make clean > /dev/null && time make -j4 > /dev/null && du -b tests/json/reflect-cpp-json-tests 

real	1m42.362s
user	6m0.853s
sys	0m21.433s
37749584	tests/json/reflect-cpp-json-tests 
```

With this we reap some of the benefits of compile-time regex, but it would be cool to move forward and be able to do compile-time parsing too. For that, we are going to need to make a lot of our internals to be constexpr (I think most of them can be constexpr as it stands). There is no reason not to do this, and would allow nice stuff to happen.

Once this PR is merged, #37 can be closed.


liuzicheng1987:
I'm amazed that things compile quicker, even though more is happening at compile time.

ChemiCalChems:
@liuzicheng1987 `std::regex` is notoriously for being incredibly slow both at compile-time and at run-time. Hana showed some data regarding this in her CTRE talks. Quite astounding it's so bad.

Issue #41, opened 2023-12-27
Missing header includes in some files


GregTheMadMonk:
My example program and compile log (**reflect-cpp** is cloned from **main**):
```
[greg@greg-spc-arch build]$ bat ../CMakeLists.txt ../main.cc
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: ../CMakeLists.txt
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ cmake_minimum_required( VERSION 3.28 )
   2   │
   3   │ project( TestRefl )
   4   │
   5   │ set( CMAKE_CXX_STANDARD 20 )
   6   │ add_subdirectory( reflect-cpp )
   7   │ set( REFLECTCPP_JSON ON )
   8   │
   9   │ add_executable( main main.cc )
  10   │ target_link_libraries( main PRIVATE reflectcpp )
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: ../main.cc
───────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ #include <rfl/json.hpp>
   2   │ #include <rfl.hpp>
   3   │
   4   │ #include <iostream>
   5   │
   6   │ struct S { int a; float b; };
   7   │
   8   │ int main() {
   9   │     S s{};
  10   │     std::cout << rfl::json::write(s) << '\n';
  11   │ }
───────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[greg@greg-spc-arch build]$ cmake .. && make
-- The C compiler identification is GNU 13.2.1
-- The CXX compiler identification is GNU 13.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done (1.5s)
-- Generating done (0.0s)
[...]

PR #44, opened 2023-12-28
Minimal fix to #41

GregTheMadMonk:
Fixes the build errors that have appeared in #41 and them only.


liuzicheng1987:
Thanks for the issue and for the fix!

PR #47, opened 2024-01-09
Adds r-value overloads for monadic operations in rfl::Result

Onlynagesha:
Changes:
1. Adds r-value overloads for monadic operations in `rfl::Result`, i.e. `rfl::Result<T>::and_then(const F&) &&` and  `rfl::Result<T>::and_then(const F&) const &&` etc.
2. Adds `operator->` for `rfl::Result` and `rfl::Validator`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 # Prerequisites
 *.d
 
+# Temporary files created by IDE or VSCode
+**/.cache
+**/.idea
+
 # Compiled Object files
 *.slo
 *.lo
--- a/include/rfl/Flatten.hpp
+++ b/include/rfl/Flatten.hpp
@@ -15,6 +15,9 @@ struct Flatten {
   /// The underlying type.
   using Type = std::remove_cvref_t<T>;
 
+  /// Default constructor with the underlying object value-initialized
+  Flatten() = default;
+
   Flatten(const Type& _value) : value_(_value) {}
 
   Flatten(Type&& _value) noexcept : value_(std::forward<Type>(_value)) {}
@@ -38,6 +41,21 @@ struct Flatten {
   Flatten(U&& _value) : value_(_value) {}
 
   ~Flatten() = default;
+
+  /// Returns the underlying object.
+  Type& operator*() { return value_; }
+
+  /// Returns the underlying object.
+  const Type& operator*() const { return value_; }
+
+  /// Returns the underlying object.
+  Type* operator->() { return &value_; }
+
+  /// Returns the underlying object.
+  const Type* operator->() const { return &value_; }
+
+  /// Returns the underlying object.
+  Type& get() { return value_; }
 
   /// Returns the underlying object.
   const Type& get() const { return value_; }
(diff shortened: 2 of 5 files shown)


liuzicheng1987:
@Onlynagesha , thanks for the contribution. I have a question, though: Could you point me to something that didn't work before the r-value overloads that works now?

As you can see, most of our tests involve items that cannot be copied, such as `std::unique_ptr` or `rfl::Box`. If the results weren't moved already, then most of our tests wouldn't compile. Could you explain what problem you are solving with the r-value overloads? Could you maybe provide a test that doesn't compile without the r-value overloads, but compiles once you do add them?

(I am not trying to be snarky or anything, I am honestly curious.)

Onlynagesha:
@liuzicheng1987 The issue was found when I attempted to use `rfl::Result` as an alternative to C++23 `std::expected`. As shown in the example above, it's impossible to modify `foo` inside the lambda function via a left-reference before this commit, which is inconsistent to the behavior of `std::expected`.
```cpp
#include <chrono>
#include <expected>
#include <fmt/color.h>
#include <iostream>
#include <rfl/Result.hpp>

struct VeryLarge {
  static inline auto default_ctor_count = 0;

  VeryLarge() : value(std::to_string(default_ctor_count++)) {
    std::cout << "Default constructor: VeryLarge()\n";
  }
  VeryLarge(VeryLarge&& other) noexcept : value(std::move(other.value)) {
    std::cout << fmt::format(fg(fmt::color::yellow), "Move constructor: VeryLarge(VeryLarge&&)\n");
    other.value = "(Moved away)";
  }
  VeryLarge(const VeryLarge& other) : value(other.value) {
    std::cout << fmt::format(fg(fmt::color::orange), "Copy constructor: VeryLarge(const VeryLarge&)\n");
  }

  std::string value;
};

// Expects consistent behavior.
template <class T>
using ResultType = rfl::Result<T>;
// using ResultType = std::expected<T, rfl::Error>;

auto some_func() -> ResultType<VeryLarge> {
  return VeryLarge{};
}

int main() {
  auto foo = some_func();
  std::cout << fmt::format("Before foo.and_then(): foo.value = {:?}\n", (*foo).value);
  // Here we attempt to modify the contents inside foo via a left-reference.
  foo.and_then([](VeryLarge& obj) -> ResultType<rfl::Nothing> {
    obj.value += "-modified"; // Modification here
    std::cout << fmt::format("foo.and_then(): obj.value: {:?}\n", obj.value);
    return rfl::Nothing{};
  });
  // We expect that the modification above is represented here.
  std::cout << fmt::format("After foo.and_then(): foo.value: {:?}\n", (*foo).value);

  // Still, for right-values, copy constructor shall not be triggered.
  some_func().and_then([](VeryLarge obj) -> ResultType<rfl::Nothing> {
    std::cout << fmt::format("some_func().and_then(): obj.value: {:?}\n", obj.value);
    return rfl::Nothing{};
  });

  return 0;
}
```

liuzicheng1987:
Hi @Onlynagesha , the compilation here still fails...could you take a look?

Issue #48, opened 2024-01-12
Compilation error for clang 16.0.6 on MacOs 14.2.1


melton1968:
I get the following error when trying to build the tests using clang 16.0.6 on MacOs 14.2.1. `main`, `v0.6.0`, `v0.5.0`, `v0.4.0` and `v0.3.0` all exhibit the same issue.

Am I missing something in the configuration or compiler? The details of the clang invocation are included below.

```
In file included from reflect-cpp/tests/json/test_all_of.cpp:4:
In file included from reflect-cpp/include/rfl/json.hpp:4:
In file included from reflect-cpp/include/rfl/json/Parser.hpp:6:
In file included from reflect-cpp/include/rfl/json/../parsing/Parser.hpp:8:
In file included from reflect-cpp/include/rfl/json/../parsing/Parser_default.hpp:10:
In file included from reflect-cpp/include/rfl/json/../parsing/../internal/enums/StringConverter.hpp:11:
reflect-cpp/include/rfl/json/../parsing/../internal/enums/get_enum_names.hpp:136:51: error: 
      non-type template argument is not a constant expression
  using EmptyNames = Names<EnumType, rfl::Literal<"">, 0>;
```

```
/opt/local/bin/clang++-mp-16  -Ireflect-cpp/include -std=c++20 -Wall -Wno-sign-compare -Wno-missing-braces -Wno-psabi -pthread -fno-strict-aliasing -fwrapv -O2 -ftemplate-backtrace-limit=0 -fsanitize=undefined -O3 -DNDEBUG -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk -std=gnu++20 -MD -MT tests/json/CMakeFiles/reflect-cpp-json-tests.dir/test_anonymous_fields.cpp.o -MF CMakeFiles/reflect-cpp-json-tests.dir/test_anonymous_fields.cpp.o.d -o CMakeFiles/reflect-cpp-json-tests.dir/test_anonymous_fields.cpp.o -c reflect-cpp/tests/json/test_anonymous_fields.cpp
```

melton1968:
I've narrowed the issue down to the following. The expression `StringLiteral("")` is not `constexpr` on MacOs. The error message indicates that a subobject is not being initialized.

```
#include <algorithm>
#include <array>
#include <string>
#include <string_view>
#include <iostream>

template <size_t N>
[...]

PR #49, opened 2024-01-12
Two tiny changes to fix compilation errors on the clang / Dawwin platform.

melton1968:
These are two tiny changes that fix compilation errors on the clang / Darwin platform.

1. Add default initialization to `arr_` in StringLiteral.hpp #48 
2. Add include for `<string>` in Result.hpp


liuzicheng1987:
I‘m on the road at the moment. Will merge as soon as I get to my destination. Thank you for figuring this out.

Issue #56, opened 2024-02-06
get_field_name_str_view() doesn't work with clang on windows


grandseiken:
The following snippet fails to compile under `clang-cl` version 17.0.6 and Visual Studio 17.8.5 with `/std:c++latest`.

```
#include <rfl.hpp>

struct test {
  bool x;
  bool y;
};

void f() {
  for (const auto& f : rfl::fields<test>()) {}
}
```

The same snippet compiles successfully when building with the microsoft compiler, or if I remove one of the fields in the struct.

The compilation error seems to be some sort of static_assert about duplicate string literals:

```
In file included from tools/test.cc:1:
In file included from external/_main~_repo_rules~reflect_cpp/include\rfl.hpp:12:
In file included from external/_main~_repo_rules~reflect_cpp/include/rfl/Attribute.hpp:10:
external/_main~_repo_rules~reflect_cpp/include/rfl/Literal.hpp(295,17): error: static assertion failed due to requirement '!has_duplicates()': Duplicate strings are not allowed in a Literal.
  295 |   static_assert(!has_duplicates(),
      |                 ^~~~~~~~~~~~~~~~~
external/_main~_repo_rules~reflect_cpp/include/rfl/internal/../internal/get_field_names.hpp(97,15): note: in instantiation of template class 'rfl::Literal<internal::StringLiteral<22>{{{103, 101, 116, 95, 102, 105, 101, 108, 100, 95, ...}}}, internal::StringLiteral<22>{{{103, 101, 116, 95, 102, 105, 101, 108, 100, 95, ...}}}>' requested here
   97 |   return rfl::Literal<_names1..., _names2...>::template from_value<0>();
      |               ^
external/_main~_repo_rules~reflect_cpp/include/rfl/internal/../internal/get_field_names.hpp(105,12): note: in instantiation of function template specialization 'rfl::internal::concat_two_literals<internal::StringLiteral<22>{{{103, 101, 116, 95, 102, 105, 101, 108, 100, 95, ...}}}, internal::StringLiteral<22>{{{103, 101, 116, 95, 102, 105, 101, 108, 100, 95, ...}}}>' requested here
  105 |     return concat_two_literals(_head, concat_literals(_tail...));
      |            ^
[...]

PR #57, opened 2024-02-06
Make library work on clang on windows

grandseiken:
This fixes https://github.com/getml/reflect-cpp/issues/56.

As described in that issue, the `function_name()` obtained from `std::source_location` on clang on windows does not include template parameters, which means the tricks used by the library to determine names automatically just don't work. However, `__PRETTY_FUNCTION__` does contain the template parameters and gives the exact string we wanted.

This works whether running clang directly on Windows or whether via `clang-cl`, since both `__clang__` and `_MSC_VER` are defined in either case.

I did a few tests of some basic functionality of the library and everything seemed to work as expected now.

### DIFF
--- a/include/rfl/internal/enums/get_enum_names.hpp
+++ b/include/rfl/internal/enums/get_enum_names.hpp
@@ -41,8 +41,12 @@ namespace enums {
 
 template <auto e>
 consteval auto get_enum_name_str_view() {
+#if defined(__clang__) && defined(_MSC_VER)
+  const auto func_name = std::string_view{__PRETTY_FUNCTION__};
+#else
   const auto func_name =
       std::string_view{std::source_location::current().function_name()};
+#endif
 #if defined(__clang__)
   const auto split = func_name.substr(0, func_name.size() - 1);
   return split.substr(split.find("e = ") + 4);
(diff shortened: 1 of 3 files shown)


grandseiken:
I managed to run the tests successfully as well after installing cmake and the ClangCl build tools for Visual Studio:

```
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DREFLECTCPP_BUILD_TESTS=ON -G"Visual Studio 17 2022" -T ClangCl
cmake --build build --config Release -j 4
./build/tests/json/Release/reflect-cpp-json-tests.exe
```

liuzicheng1987:
@grandseiken , first of all, thank you so much for the PR. 

I have one minor request though:

There are two more places where source_location::function_name is used:

https://github.com/getml/reflect-cpp/blob/main/include/rfl/internal/get_type_name.hpp

https://github.com/getml/reflect-cpp/blob/main/include/rfl/internal/enums/get_enum_names.hpp

Could you implement your proposed changes in these functions as well (I am fairly sure a simple copy + paste would do the trick) and then also make sure that our tests compile and run through on Clang on Windows? (JSON-only is fine.)

I will then have to figure out a way to integrate Clang on Windows into our testing pipeline.

liuzicheng1987:
@grandseiken , thank you so much for this PR.

Issue #58, opened 2024-02-07
Convenient ways to iterate over fields at compile-time


grandseiken:
The library claims to support reflective programming in general, but found that I had to do a bit more work to really make it convenient to iterate over fields of a struct with type information and field names, e.g.

```cpp
template <typename T, typename F>
void for_each_field(T&& value, F&& f) {
  std::apply(
      [&f]<typename... Fields>(Fields&&... fields) {
        ((f.template operator()<std::remove_cvref_t<Fields>::name_>(
             *std::forward<Fields>(fields)())),
         ...);
      },
      rfl::to_view(value).fields());
}

// example usage
auto print_ints = 
  []<auto FieldName, typename T>(const T& x) {
    if constexpr(std::is_same_v<T, int>) {
      std::cout << FieldName.string_view() << ": " << x << "\n";
    }
  };

struct foo { int x; float y; int z; };
for_each_field(foo{ 1, 2.0f, 3 }, print_ints);
```

Did I miss something that would have made this easier? Do you think helper functions like this would make sense to include in the library, or maybe in example documentation?

I'm also not sure if I missed a better way to get the name of an `rfl::Field` as a compile-time value, I had to use the internal `name_` member above.

liuzicheng1987:
I think a helper function like this would be a good addition, but there are a couple of things I would do differently:

1) I would design it as a method called `.apply(...)` on the NamedTuple.

https://github.com/getml/reflect-cpp/blob/main/include/rfl/NamedTuple.hpp

(`rfl::view(...)` just returns a named tuple with pointers to the original fields.)

`.apply(...)` should also have a const version.

2) I would design it in such a way that the function F expects an actual field. So in your example I would rewrite `print_ints` as follows:

```cpp
auto print_ints = 
  [](const auto& field) {
    if constexpr(std::is_same_v<typename Field::Type, int>) {
      std::cout << field.name() << ": " << *field.value() << "\n";
    }
  };

// Use as follows:
rfl::view(foo{ 1, 2.0f, 3 }).apply(print_ints);
```

(`field.name()` will have to be added as well. I already have an implementation, but it is in a feature branch. It works as follows:

```cpp
struct Field {
   ...
    static std::string name() { return Name().str(); } 
}
```

https://github.com/getml/reflect-cpp/blob/main/include/rfl/Field.hpp

This works because Field::Name is a Literal and literals have this method:

https://github.com/getml/reflect-cpp/blob/main/include/rfl/Literal.hpp

[...]

PR #59, opened 2024-02-07
add rfl::NamedTuple::apply() for convenient compile-time iteration

grandseiken:
Adds `rfl::NamedTuple::apply()` as proposed in https://github.com/getml/reflect-cpp/issues/58.

I made one small deviation from the proposal which was to implement `rfl::Field::name()` by way of its `rfl::internal::StringLiteral` rather than `rfl::Literal` so that it can be a compile-time constant, which I think is important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ for (const auto& f : rfl::fields<Person>()) {
 }
 ```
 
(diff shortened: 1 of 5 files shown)


liuzicheng1987:
I think the compile-time name is a great idea!

PR #63, opened 2024-02-13
fix: handle wstrings being different lengths and constructed correctly

Lazrius:
Hiya. I made a mistake when doing the original wstring parser. It would crash when trying to construct a wstring for reading due to the parameters being inverted. I tested it on MSVC which worked fine, but when setting it up on Clang it crashed.


liuzicheng1987:
Ok. Thanks for the contribution!

PR #66, opened 2024-02-22
Add helpers for inspecting enum types

grandseiken:
In the same spirit as https://github.com/getml/reflect-cpp/pull/59, I'd like reflect-cpp to expose general-purpose enum reflection utilities (since it already knows how to do this internally), without having to depend on `rfl::internal` or implement things myself.

This PR adds four functions in a new `rfl/enums.hpp` file:

* `rfl::enum_to_string(enum)` and `rfl::string_to_enum<EnumType>(string)` are simple wrappers around the existing internal functions.
* `rfl::get_enumerators<EnumType>()` gives back a named tuple of enumerator names/values, while `rfl::get_enumerator_array<EnumType>` gives back the same data as an `std::array` of pairs (the main advantage of this one being that it can be `constexpr`).

Example:

```
enum class Color { red, green, blue };
get_enumerators<EnumType>() == NamedTuple{make_field<"red">(Color::red), make_field<"blue">(Color::blue), make_field<"green">(Color::green)}
get_enumerator_array<EnumType>() == std::array{std::make_pair("red", Color::red), std::make_pair("green", Color::green), std::make_pair("blue", Color::blue)}
```

The tests pass along with some new ones. If you like these changes I'll add some documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,9 @@ This results in the following JSON string:
 {"pos_x":2.0,"pos_y":3.0,"shape":"square","color":"red|blue"}
 ```
 
+You can also directly convert between enumerator values and strings with `rfl::enum_to_string()` and `rfl::string_to_enum()`, or
+obtain list of enumerator name and value pairs with `rfl::get_enumerators<EnumType>()` or `rfl::get_enumerator_array<EnumType>()`.
+
 ## Algebraic data types
 
 reflect-cpp supports Pydantic-style tagged unions, which allow you to form algebraic data types:
--- a/docs/enums.md
+++ b/docs/enums.md
@@ -146,3 +146,39 @@ This will be represented as follows:
 
 This works, because 16 + 256 + 512 + 1024 + 8192 = 10000. Flag enums are *always* represented in terms of 2^N-numbers.
 
+## General-purpose enumeration utilities
+
+reflect-cpp also allows you to directly convert between enumeration values and strings:
+
+```cpp
+enum class Color { red, green, blue };
+
+auto name = rfl::enum_to_string(Color::red);       // "red"
+auto value = rfl::string_to_enum<Color>("red");    // result containing Color::red
+auto value = rfl::string_to_enum<Color>("greem");  // error result
+```
+
+This works with normal and flag enums, and behaves just like serialization of types containing enumerations as described above.
+
+You can also inspect the defined enumerators of an enum type (including at compile-time):
+
+```cpp
+enum class Color { red, green, blue };
+
+// This produces a named tuple where the keys are "red", "green", and "blue", with corresponding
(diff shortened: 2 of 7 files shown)


liuzicheng1987:
I think it looks great.

My only question is, should the named tuple and array contain the enum values themselves or the underlying value? I think if you were to serialize the resulting names tuple, it would read “yellow”: “yellow”, “green”: “green”, etc, which is not very informative. But if you were to cast them to the underlying type, you would get a very helpful mapping.

But if you have good arguments why your current approach is more useful, I am open to discussion.

liuzicheng1987:
But certainly, I would like to add these changes, so please do add documentation.

liuzicheng1987:
@grandseiken , I can also write a parser that would allow us to serialize std::string_view. That way we could serialize the enumerator array like this:

```cpp
std::cout << rfl::yaml::write(get_enumerator_array<EnumType>()) << std::endl;
```

If we then use the underlying type, like I suggested, this would give you a very readable overview of the enum.

grandseiken:
> My only question is, should the named tuple and array contain the enum values themselves or the underlying value? I think if you were to serialize the resulting names tuple, it would read “yellow”: “yellow”, “green”: “green”, etc, which is not very informative. But if you were to cast them to the underlying type, you would get a very helpful mapping.

I expected that this functionality would be primarily useful for generic reflection use-cases that do not involve serializing the results. (As an example, suppose you want to run some internal logic for each case of an enum using `rfl::get_enumerators<MyEnum>().apply(f)`.) In such cases, having the values be those of the actual enum type is more convenient and more type-safe.

I didn't really consider wanting to serialize the result. Since the data is completely static that seems of limited utility, but I guess there could be odd use cases like for self-describing serialization formats, schema generation or help text.

I suppose either choice is at least somewhat reasonable: the names map to enums values from the perspective of _users_ of the enum, but inside the _definition_ of the enum you do write values of the underlying type.

If you feel strongly I can change it, or maybe it'd be better to just support both with something like `rfl::get_enumerators<EnumType>()` and `rfl::get_underlying_enumerators<EnumType>()`?

liuzicheng1987:
Sure, let’s just support both. The way I see it, it’s just one more .apply(…) to get the underlying enumerators

liuzicheng1987:
@grandseiken , sorry, fat fingers…how should be proceed here? Do you want to add the documentation first or should I merge the PR and we will add it later?

grandseiken:
Sorry for the delay, I forgot about this. I'll try to get it sorted tomorrow.

liuzicheng1987:
No problem

grandseiken:
Add documentation and the underlying versions of the functions.

liuzicheng1987:
Merged. 

@grandseiken , thank you so much for the contribution!

PR #74, opened 2024-03-21
undo change of parameter type in `NamedTuple::apply`

grandseiken:
58f1e41c830d2e598415e480939e38df3a76f785 changed the parameter type of the named tuple `apply` method from `F&&` to `const F&`, but this prevents it from being used with non-const objects. There is certainly no reason to disallow this (my fault for not adding a test).

### DIFF
--- a/include/rfl/NamedTuple.hpp
+++ b/include/rfl/NamedTuple.hpp
@@ -602,7 +602,7 @@ class NamedTuple<> {
 
   /// Does nothing at all.
   template <typename F>
-  void apply(const F& _f) const {}
+  void apply(F&& _f) const {}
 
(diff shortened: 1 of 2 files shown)


liuzicheng1987:
Just to be clear...you want to pass an object to apply(...) that has an operator(...) which then has a side-effect on the object?

I am very much into functional programming and I personally think doing something like this is a code smell, but I guess you are right...there is no reason to disallow this. C++ is about freedom and letting people make their own design choices.

grandseiken:
`apply()` returns `void`, so the object passed _must_ represent a side-effect of some sort. Whether the side effect happens to be on the object itself or some other object know to that object is immaterial.

PR #76, opened 2024-04-01
Use BUILD_SHARED_LIBS as default value for REFLECTCPP_BUILD_SHARED

hlewin:
`BUILD_SHARED_LIBS` is the standard cmake variable to indicate if output should be shared libraries, so this should be used to streamline standalone builds.  To not break existing integrations and to make inclusion (eg as submodule) easier this patch leaves `REFLECTCPP_BUILD_SHARED` in place but default-initializes it with the value of `BUILD_SHARED_LIBS`.


liuzicheng1987:
Great input! Thanks for the contribution!

PR #77, opened 2024-04-01
Add cmake config module

hlewin:
This makes reflectcpp installable by cmake and adds a config module, so that it can be used like any other cmake package

```
find_package(reflectcpp REQUIRED)

target_link_libraries(MY_TARGET reflectcpp::reflectcpp)
```


liuzicheng1987:
Again, thank you so much!

Issue #82, opened 2024-04-09
Error when struct has a base class


lixin-wei:
The following code will fail to compile:

```cpp
struct EmptyBase {};
struct DerivedStruct : public EmptyBase {
  int x;
};

DerivedStruct derived;
rfl::to_view(derived);
```

Because the auto generated default constructor of `DerivedStruct` is `DerivedStruct(EmptyBase, int)`. Which has 2 params.

So `has_n_fields<DerivedStruct, 2>` will hold, then `auto &[f1, f2] = derived` will fail.

I've searched around and just find a non-standard gcc traits called [direct_base](https://gcc.gnu.org/onlinedocs/gcc-4.7.2/libstdc++/api/a00909.html) which might help.

Do you guys have any better idea?

schaumb:
I have a compiler-independent solution for that.

When the detection happens (CountFieldHelper), it should be a specific ubiq type, which can be convertible to only the class's base class. 
https://github.com/vizzuhq/vizzu-lib/blob/main/src/base/refl/auto_struct.h#L29
With this ubiq, when the wrong field count is first calculated, it should be checked that the aggregated initialization first K "member" can be a parent class of the original. https://github.com/vizzuhq/vizzu-lib/blob/main/src/base/refl/auto_struct.h#L60-L64

General implementation does not exist yet as far as I know. 
https://github.com/boostorg/pfr/issues/56#issuecomment-1824106120


Special cases
---

There are some special classes where it doesn't work when the parent class is a member too.

like the following:

```cpp
struct Base {};

struct Derived: Base {
  Base as_member_too; // recognized as base class
};
```

A method should exist that can filter these unconventional cases with some compilation time structures, like a compile time set that tracks the already recognized base classes.

--- 

Other special case when the grandparent is the member:

```cpp
struct BaseBase{};
struct Base: BaseBase{};

struct Derived: Base {
  BaseBase memb; 
};
```

This can be eliminated with a recursive base-ubiq recognizer and a compile time set. But with the technique, this class is the same, which makes the solution unambiguous.
```cpp
struct BaseBase{};
struct Base: BaseBase{};

struct Derived: Base, BaseBase {};
```

liuzicheng1987:
@schaumb , I will take a closer look at your solution and add it to the library.

lixin-wei:
@liuzicheng1987 Do you start coding? If not I'm willing to have a try.

liuzicheng1987:
[...]

PR #83, opened 2024-04-10
Support one-level inheritance cases

lixin-wei:
Resolves <https://github.com/getml/reflect-cpp/issues/82>

Please squash merge this PR.

### DIFF
--- a/include/rfl/internal/num_fields.hpp
+++ b/include/rfl/internal/num_fields.hpp
@@ -29,6 +29,7 @@ the potential array as we can without missing variables in subsequent fields.
 This is the purpose of get_nested_array_size().
 */
 
+#include <algorithm>
 #include <cstddef>
 #include <type_traits>
 #include <utility>
@@ -109,31 +110,26 @@ struct CountFieldsHelper {
   }
 
   template <std::size_t size = 0>
-  static consteval std::size_t get_the_sole_nested_array_size() {
+  static consteval std::size_t get_the_sole_nested_base_field_count() {
     static_assert(size <= sizeof(T));
     if constexpr (constructible_with_nested<0, size, 0>() &&
                   !constructible_with_nested<0, size + 1, 0>()) {
       return size;
     } else {
-      return get_the_sole_nested_array_size<size + 1>();
+      return get_the_sole_nested_base_field_count<size + 1>();
     }
   }
 
   template <std::size_t n, std::size_t total_arg_num>
   static consteval bool has_n_base_param() {
-    if constexpr (n > total_arg_num) {
-      return false;
-    } else {
-      auto left = std::make_index_sequence<n>();
-      auto right = std::make_index_sequence<total_arg_num - n>();
-      return []<std::size_t... l, std::size_t... r>(std::index_sequence<l...>,
-                                                    std::index_sequence<r...>) {
-        return requires { T{any_base<T>(l)..., any(r)...}; };
-      }(left, right);
-    }
+    constexpr auto right_len = total_arg_num>=n ? total_arg_num-n : 0;
+    return []<std::size_t... l, std::size_t... r>(std::index_sequence<l...>,
+                                                  std::index_sequence<r...>) {
+      return requires { T{any_base<T>(l)..., any(r)...}; };
+    }(std::make_index_sequence<n>(), std::make_index_sequence<right_len>());
   }
 
-  template <std::size_t total_arg_num, std::size_t index>
+  template <std::size_t total_arg_num, std::size_t index = 0>
   static consteval std::size_t base_param_num() {
     if constexpr (!has_n_base_param<index + 1, total_arg_num>()) {
       return index;
@@ -158,7 +154,7 @@ struct CountFieldsHelper {
     constexpr std::size_t max_fields = count_max_fields();
     constexpr std::size_t total_args =
         constructible_no_brace_elision<0, max_fields>();
-    constexpr std::size_t base_args = base_param_num<total_args, 0>();
+    constexpr std::size_t base_args = base_param_num<total_args>();
     if constexpr (total_args == 0 && base_args == 0) {
       // Empty struct
       return 0;
@@ -166,7 +162,7 @@ struct CountFieldsHelper {
       // Special case when the derived class is empty.
       // In such cases the filed number is the fields in base class.
       // Note that there should be only one base class in this case.
-      return get_the_sole_nested_array_size();
+      return get_the_sole_nested_base_field_count();
     } else {
       return total_args - base_args;
     }
--- a/include/rfl/internal/num_fields.hpp
+++ b/include/rfl/internal/num_fields.hpp
@@ -48,6 +48,17 @@ This is the purpose of get_nested_array_size().
 namespace rfl {
 namespace internal {
 
+template <class Derived>
+struct any_base {
+  any_base(std::size_t);
+  template <class Base>
+    requires(
+        std::is_base_of_v<std::remove_cvref_t<Base>,
+                          std::remove_cv_t<Derived>> &&
+        !std::is_same_v<std::remove_cvref_t<Base>, std::remove_cv_t<Derived>>)
+  constexpr operator Base&() const noexcept;
+};
+
 struct any {
   any(std::size_t);
   template <typename T>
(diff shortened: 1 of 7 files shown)


lixin-wei:
@liuzicheng1987 Hi I'm done. Would you please have a look when you have time?

include/rfl/internal/num_fields.hpp
```diff
@@ -97,20 +108,55 @@ struct CountFieldsHelper {
...
 
   static consteval std::size_t count_fields() {
-    return count_fields_impl<0, count_max_fields()>();
+    constexpr std::size_t total_args =
+        constructable_no_brace_elision<0, count_max_fields()>();
+    constexpr std::size_t base_args = base_param_num<total_args, 1>();
+    if constexpr (base_args == total_args) {
+      // Special case when the derived class is empty.
+      // In such cases the filed number is the fields in base class.
+      return total_args;
```
schaumb: This will return with the number of the base classes and not the number of fields.

include/rfl/internal/num_fields.hpp
```diff
@@ -97,20 +108,55 @@ struct CountFieldsHelper {
...
+  template <std::size_t total_arg_num, std::size_t index>
+  static consteval std::size_t base_param_num() {
+    if constexpr (index > total_arg_num) {
+      return 0;
+    } else if constexpr (has_n_base_param<index, total_arg_num>() &&
+                         !has_n_base_param<index + 1, total_arg_num>()) {
+      return index;
+    } else {
+      return base_param_num<total_arg_num, index + 1>();
+    }
```
schaumb: this should be simplified to only 2 branches, something like this. 
```cpp
if constexpr (!has_n_param<index, .. >)
  return index - 1;
else
  return base_param_num<.., index + 1>();
```

include/rfl/internal/num_fields.hpp
```diff
@@ -97,20 +108,55 @@ struct CountFieldsHelper {
     }
   }
 
+  template <std::size_t n, std::size_t total_arg_num>
+  static consteval bool has_n_base_param() {
+    if constexpr (n > total_arg_num) {
+      return false;
+    } else {
```
schaumb: this branch is unnecessary because the `requires {}` clause will handle it.

lixin-wei: It's necessary because if `n > total_arg_num`, `std::make_index_sequence<total_arg_num - n>();` will fail  to compile.

So I added this branch to avoid instantiating it.

include/rfl/internal/num_fields.hpp
lixin-wei: Nice catch, fixed, thx.

include/rfl/internal/num_fields.hpp
lixin-wei: Fixed, thanks.

lixin-wei:
@schaumb Thanks for the review. All comments fixed/replied. Would you pls have a look again?

liuzicheng1987:
@lixin-wei , great work!

I have two very minor requests:

1. Could you place your test in a separate test and call it "test_inheritance2"? The reason I think this is a good idea is that there is already "test_inheritance" which basically tests the flip case of what you have implemented: https://github.com/lixin-wei/reflect-cpp/blob/derived_srturct/tests/json/test_inheritance.cpp I also think every test should only test one thing.

I realize that your test is a compile-time test and I think that is a very good thing. Just std::cout OK at the end for consistency.

2. The word "constructible" is spelled with an i, not an a...I realize that that doesn't make a lot of sense, but English spelling is notoriously inconsistent. So could you rename your function `constructable_no_brace_elision`?

include/rfl/internal/num_fields.hpp
schaumb: If the `base_param_num` calls only when `index + 1 <= total_arg_num`, then no needed this branch 

```cpp
if constexpr (index == total_arg_num || !has_n_base_param<index + 1, total_arg_num>()) {
 return index;
...
```

( or `!has_n_base_param<index + 1, std::max({index + 1, total_arg_num})>()` )

include/rfl/internal/num_fields.hpp
```diff
@@ -97,20 +108,68 @@ struct CountFieldsHelper {
     }
   }
 
+  template <std::size_t size = 0>
+  static consteval std::size_t get_the_sole_nested_array_size() {
```
schaumb: Wrong name: this is not a "nested array" size check.

include/rfl/internal/num_fields.hpp
```diff
@@ -97,20 +108,68 @@ struct CountFieldsHelper {
...
+    constexpr std::size_t total_args =
+        constructable_no_brace_elision<0, max_fields>();
+    constexpr std::size_t base_args = base_param_num<total_args, 0>();
+    if constexpr (total_args == 0 && base_args == 0) {
+      // Empty struct
+      return 0;
+    } else if constexpr (base_args == total_args) {
+      // Special case when the derived class is empty.
+      // In such cases the filed number is the fields in base class.
+      // Note that there should be only one base class in this case.
```
schaumb, line 195: This is only true with this implementation, but not necessarily. 

You can check which base class is 'empty' and which is not with another ubiq: 
```cpp

template <class Derived>
struct any_empty_base {
  any_empty_base(std::size_t);
  template <class Base>
    requires(
        std::is_empty_v<std::remove_cvref_t<Base>> && // new check
        std::is_base_of_v<std::remove_cvref_t<Base>,
                          std::remove_cv_t<Derived>> &&
        !std::is_same_v<std::remove_cvref_t<Base>, std::remove_cv_t<Derived>>)
  constexpr operator Base&() const noexcept;
};
```

`T(any_empty_base(Bx...), any_base(N),  any_empty_base(Ax...))`. 
If no one is creatable (N <- [1..base_args]) then it is not a structure bindable class (static_assert).

lixin-wei:
@liuzicheng1987 All done would you please have a look again? thanks for taking the time to review!
Note that please squash merge my PR because there are a lot of noisy commit like 'fix CI', 'fix coments'.

BTW, can we set the default merging strategy to 'squash merge' in our project? I've found lots of non-squash-merged commits in log history, which make the log not so clean.

include/rfl/internal/num_fields.hpp
```diff
@@ -97,20 +108,68 @@ struct CountFieldsHelper {
...
+      return get_the_sole_nested_array_size();
```
schaumb: This should be implemented with a recursion somehow because there are classes with multiple inheritances that can be used with structure bind. 

```cpp
struct B1 {};
struct B2 {};
struct B3 {};
struct M { int a; int b; };
struct D1 : B1, M, B3 {};
struct D2 : B2, D1, B3 {};

// ...
// D2 d;
// rfl::to_view(d) // recognized 3, but only 2 member
```

lixin-wei:
> but when I squash + merge, then I become the author (at least in some cases).
Really? I don't think so. Github does handle this when you click 'squash merge'. The author won't be you.

For example, [this commit](https://github.com/ray-project/ray/commit/c4b9e9ffa54a9272f3b791ab7a61fc814d4257f6) is a squash merge from [this PR](https://github.com/ray-project/ray/pull/26073). The one who squash merged the PR is not the author but I am.

lixin-wei:
> but when I squash + merge, then I become the author (at least in some cases).

@liuzicheng1987 Really? I don't think so. Github does handle this when you click 'squash merge'. The author won't be you.

For example, [this commit](https://github.com/ray-project/ray/commit/c4b9e9ffa54a9272f3b791ab7a61fc814d4257f6) is a squash merge from [this PR](https://github.com/ray-project/ray/pull/26073). The one who squash merged the PR is not the author but I am.

liuzicheng1987:
@lixin-wei , I have had this issue under some circumstances. It is possible that this only happens when I have to resolve merge conflicts.

But I am certainly not the only one who has had this problem:

https://stackoverflow.com/questions/39700184/git-squash-commits-while-retaining-authors-info

Again, the issue here might be merge conflicts...

I will certainly research this issue, because I also prefer squash + merge.

liuzicheng1987:
@lixin-wei , do you want to implement @schaumb latest input? I think the multiple inheritance edge case is a valid point.

lixin-wei:
@liuzicheng1987 Yes I'm working on it, there are so many corner cases... 😭 I'll ping you when I'm finished.

lixin-wei:
>I have had this issue under some circumstances. It is possible that this only happens when I have to resolve merge conflicts.

@liuzicheng1987 It's author's responsibility to resolve the conflicts. You shouldn't do this.

If you do help the author to resolve the conflicts. Then you do 'contributed' to this PR. There is no problem to put you into the author list. Github will show all persons contributed to the PR in the squash merge commit, like the following image:

image

liuzicheng1987:
@lixin-wei , I will just give it a try. I will just make squash + merge the default. If there are any issues along the lines I have outlined, I can always change it back.

include/rfl/internal/num_fields.hpp
lixin-wei: I found a btter solution:

```cpp
  template <std::size_t n, std::size_t total_arg_num>
  static consteval bool has_n_base_param() {
    constexpr auto right_len = total_arg_num>=n ? total_arg_num-n : 0; // here!
    return []<std::size_t... l, std::size_t... r>(std::index_sequence<l...>,
                                                  std::index_sequence<r...>) {
      return requires { T{any_base<T>(l)..., any(r)...}; };
    }(std::make_index_sequence<n>(), std::make_index_sequence<right_len>());
  }
```

include/rfl/internal/num_fields.hpp
lixin-wei: changed to `get_the_sole_nested_base_field_count`

lixin-wei:
Hi @liuzicheng1987 I solved the one-level multi inheritance with the `any_empty_base` mentioned above. Thanks @schaumb !

But the multi-level inheritance case(as the following code shows) is very tricky and I don't have a very good solution😂 Do you guys have any idea? Can we merge this first and open another PR to tackle this hard case?

```cpp
struct B1 {};
struct B2 {};
struct B3 {};
struct M { int a; int b; };
struct D1 : B1, M, B3 {};
struct D2 : B2, D1, B3 {};
```

liuzicheng1987:
@lixin-wei , I think so, too...I mean this is a bit of an edge case anyway.  The possible use cases for a multiple inheritance structure like this are fairly limited anyway...

liuzicheng1987:
@lixin-wei , seems you were right. I squashed + merged this and you are now the author of the squashed merge in the git history.

lixin-wei:
@liuzicheng1987 I also remcomend you to install [this plugin](https://chromewebstore.google.com/detail/squashed-merge-message/kgabfelbdpeifcemndligpdfddhbbmfm), which will automatically set the commit message from PR's description. The default squash commit message of GitHub is very ugly.

liuzicheng1987:
@lixin-wei , cool, I didn't know that one!

schaumb:
I created a multi-level CountFieldsHelper here:
https://godbolt.org/z/bqsq5Wv7s

lixin-wei:
@schaumb That's awesome. I've learned a lot from your code.

liuzicheng1987:
@lixin-wei , if you want to open another PR, feel free to do so. I would love to have this.

PR #119, opened 2024-06-12
Separate third party libraries; resolves #79

### LINKED ISSUE #79: Separate yyjson and ctre?
Hello!

I was thinking about if it would be possible to separate reflect-cpp from yyjson, ie not to bundle yyjson per se but to do so only if requested by cmake options. The alternative would be to rely on vcpkg or to rely on a system-wide installation of yyjson.

This would have the following benefits:
- reflect-cpp could be used as pure reflection library
- yyjson would not have to be maintained (updated) internally
- For the case of a system-wide installation of yyjson and reflect-cpp there would be no file clashes

The same argument applies to the `ctre.hpp` shipped with reflect-cpp. This should also be separated for the same reasons.

Any thoughts on this?

liuzicheng1987:


CMakeLists.txt
```diff
@@ -29,15 +31,32 @@ project(reflectcpp)
...
-    add_library(reflectcpp STATIC src/yyjson.c)
-endif ()
+if(REFLECTCPP_USE_BUNDLED_DEPENDENCIES)
+    if (REFLECTCPP_BUILD_SHARED)
+        add_library(reflectcpp SHARED src/yyjson.c)
+    else()
+        add_library(reflectcpp STATIC src/yyjson.c)
+    endif()
+else()
+    if (REFLECTCPP_BUILD_SHARED)
```
Urfoex, line 41: Probably instead of `SHARED` or `STATIC`, `INTERFACE` would be the choice for Header-Only-Lib
https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

CMakeLists.txt
```diff
@@ -29,15 +31,32 @@ project(reflectcpp)
...
+        add_library(reflectcpp SHARED)
+    else()
+        add_library(reflectcpp STATIC)
+    endif()
+endif()
 
+set_target_properties(reflectcpp PROPERTIES LINKER_LANGUAGE CXX)
```
Urfoex, line 48: Instead, probably could use the `LANGUAGES` section of `project`
https://cmake.org/cmake/help/latest/command/project.html

But setting it just to `CXX`, as this is a C++ project, gives trouble with the yyjson.c .
I guess it could be a C++ wrapper around it. 
Maybe like 
```cpp
// yyjson.cpp
extern C {
  #include "yyjson.c"
}
```

Quick test shows, that at least clangd is happy with it, and uses the right compiler and flags.

CMakeLists.txt
```diff
@@ -29,15 +31,32 @@ project(reflectcpp)
...
 target_compile_features(reflectcpp PUBLIC cxx_std_20)
 
-target_include_directories(reflectcpp PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include> )
+if(REFLECTCPP_USE_BUNDLED_DEPENDENCIES)
+    target_include_directories(reflectcpp PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:include> $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/rfl/thirdparty>)
```
Urfoex, line 52: If the `thirdparty` folder as set as extra include here, there should be no need to have the checks in the header files. I guess. Compiler should use the first header it finds, so if `thirdparty` is first, it should be okay.

PR #124, opened 2024-06-26
Feature/compatibilities gcc14 clang14 vcpkg

Urfoex:
Compatibility GCC14:
- upgrade simdjson (https://github.com/simdjson/simdjson/issues/2116, https://github.com/simdjson/simdjson/issues/2175)
  
Compatibility clang14:
- wrap source_location and move unused includes
- use constructor via {}
- <=> in three ways
-  add missing typename

Compatibility vcpkg (https://github.com/microsoft/vcpkg/pull/38899):
- be an interface library if being header-only / without bundled dependencies


liuzicheng1987:
@Urfoex , great work!

Just one minor question: Why did you remove the `OR REFLECTCPP_BUILD_BENCHMARKS` in the CMake?

PR #133, opened 2024-06-30
Use -O3

jkeiser:
Use -O3 for more optimizations. simdjson in particular sees benefits when running at this optimization level.

DRAFT: I haven't got this running yet, so I'm abusing CI for this small change. I'll get it up locally too :)

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -1,9 +1,9 @@
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -O2")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -O3")
 
 if (MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std:c++20 -O2")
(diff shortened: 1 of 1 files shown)


benchmarks/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -O2")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -O3")
 
 if (MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std:c++20 -O2")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std:c++20 -O3")
```
StephanTLavavej, line 4: This is wrong - MSVC does not support `-O3`. See [`/O` options (Optimize code)](https://learn.microsoft.com/en-us/cpp/build/reference/o-options-optimize-code?view=msvc-170), `-O2` is the highest option that MSVC supports.

jkeiser:
Closing. Don't think it helps.

PR #134, opened 2024-06-30
simdjson inlining and idiomaticity in benchmark

jkeiser:
This doesn't really make simdjson any faster than BSON or other formats, but it does bring it back in line with my expectations around performance :) After this, canada.json is the same or faster than yyjson on my machine (but only by pennies!), and remains faster on the other benchmarks.

The two changes that made the difference:

1. Inlining the deserialization methods. On Demand is explicitly designed to take advantage of inlining.
2. Reading the coordinate tuple [1,2] with an iterator instead of indexing it as an array.

I also removed some unnecessary boilerplate .value() and .get_array/object() methods.

NOTE: simdjson could get some small perf gains if `rfl::Literal` took a string_view constructor; the fact that it takes a `std::string` means we're forced to allocate every time we see a `type`, when all it really needs to do is compare the string to a static list and put the integer index in there. This seems like it would a generally positive change to reflect-cpp anyway :)

### DIFF
--- a/benchmarks/json/canada.cpp
+++ b/benchmarks/json/canada.cpp
@@ -220,22 +220,26 @@ simdjson_inline std::vector<std::vector<std::tuple<double, double>>> simdjson_to
   return coordinates;
 }
 
+simdjson_inline std::string simdjson_to_string(simdjson::ondemand::value _val) {
+  return std::string(std::string_view(_val));
+}
+
 simdjson_inline Property simdjson_to_property(simdjson::ondemand::object _val) {
   Property property;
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
@jkeiser thank you so much.

I will also add support for constructing a literal using std::string_view.

jkeiser:
I wanted to point out here what I posted on the Reddit thread: I haven't run this on msvc. It might still be slower there, or might be fixed by the incoming (which is a pretty huge change from the optimizer's point of view, even if none of the C++ changes).

PR #137, opened 2024-07-03
Fix typo in custom parser docs

tim-lyon:
For the example to compile, the `ProcessorsType` needs to be passed though to the custom perser's template params?

### DIFF
--- a/docs/custom_parser.md
+++ b/docs/custom_parser.md
@@ -78,7 +78,7 @@ You then implement the custom parser:
 namespace rfl::parsing {
 
 template <class ReaderType, class WriterType, class ProcessorsType>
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
Thanks for the contribution. You are right, of course.

tim-lyon:
Thanks for merging @liuzicheng1987!
I also suspect the documentation on [anonymous fields](https://github.com/getml/reflect-cpp/blob/main/docs/anonymous_fields.md) might be out of date, as it appears the default behaviour now picks up the field names? Is it worth opening an issue for? My follow-on question is, with the new behaviour, is there a way to _force_ a field to become anonymous? Thanks, loving the library so far!

PR #139, opened 2024-07-05
Fix typo in readme

tim-lyon:
A tiny typo, but because the documentation standard is otherwise so high, I thought it worth fixing :)


liuzicheng1987:
Thanks!

Issue #141, opened 2024-07-05
potentially simpler way to do custom parsing


BjornSAIM:
Currently you have to define a new aggregate type and a parser specialization to achieve this. After fiddling around for a bit I was able to get this alternative to work using concepts:

```
// defined in the library
template <T>
struct Reflector;

// defined by the user
template <>
struct Reflector<Vector3> {
    struct refl_type {
        float x = {}, y = {}, z = {};
    };
    static constexpr Vector3 to(const refl_type& v) noexcept
    {
        return { v.x, v.y, v.z };
    }

    static constexpr refl_type from(const Vector3& v)
    {
        return { v.x, v.y, v.z };
    }
};
```

It's conceptually similar to what is happening now, but to me this seems more straightforward, it feels a bit like implementing a serialize function pair. Ofcourse that's just my opinion though.

I also had a question. Is it expected behavior that the library can't handle structs that have all public fields but explicitly defined constructors?

liuzicheng1987:
@BjornSAIM, this seems great, I would like to see how it is implemented, though. Does it work for formats other than JSON? How do you handle processors? Could you share the code or maybe open a PR that we could discuss further? 

As for your question…it’s not really intended behaviour, it’s more of a necessary evil. It’s also documented, refer to the very bottom of this page:

https://github.com/getml/reflect-cpp/blob/main/docs/structs.md

liuzicheng1987:
@BjornSAIM after giving it some more thought, I have now figured out how this works. I think it is a great idea. A PR would be very welcome.

BjornSAIM:
> @BjornSAIM, this seems great, I would like to see how it is implemented, though. Does it work for formats other than JSON? How do you handle processors? Could you share the code or maybe open a PR that we could discuss further?
> 
> As for your question…it’s not really intended behaviour, it’s more of a necessary evil. It’s also documented, refer to the very bottom of this page:
> 
> https://github.com/getml/reflect-cpp/blob/main/docs/structs.md

That's what I expected, but in a case like this:

```
struct Vector3 {
    float x = {}, y = {} ,z = {};
    constexpr Vector3() = default;
    constexpr Vector3(float x, float y, float z)
        : x(x), y(y), z(z)
    {
    }
...
};
```

I would expect this to match the criterium. I think it might be possible to soften the constrain from std::is_aggregate_v to is constructible in this way using concepts.

liuzicheng1987:
[...]

PR #142, opened 2024-07-06
added a new method of implementing a custom parser using a Reflector …

BjornSAIM:
see: https://github.com/getml/reflect-cpp/issues/141

### DIFF
--- a/docs/custom_parser.md
+++ b/docs/custom_parser.md
@@ -1,7 +1,28 @@
 # Custom parsers
 
 If you absolutely do not want to make any changes to your original classes whatsoever,
-you can implement a custom parser using `rfl::parsing::CustomParser`.
+You can create a Reflector template specialization for your type:
+
+```cpp
+namespace rfl {
+template <>
+struct Reflector<Person> {
+  struct ReflType {
+    std::string first_name;
+    std::string last_name;
+  };
+  static constexpr Person to(const ReflType& v) noexcept {
+    return {v.first_name, v.last_name};
+  }
+
+  static constexpr ReflType from(const Person& v) {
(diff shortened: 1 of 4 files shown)


liuzicheng1987:
A couple of quick comments:

1. I don't see any implementation of the reflector itself, only a concept...if people are going to use this, wouldn't you have to provide an implementation as well?
2. Could you add a test (only for JSON, there is no need to copy + paste it into all the formats)?
3. Could you also add some documentation on how to use your class? (If it's too much of a hassle, I can do it as well, no problem).
4. Very minor thing: In this library, we generally use CamelCase for types and classes...so could you rename "refl_type" into something like "ReflType"?

liuzicheng1987:
But it is such a great idea, thank you so much!

BjornSAIM:
> A couple of quick comments:
> 
> 1. Could you add a test (only for JSON, there is no need to copy + paste it into all the formats)?
> 2. Could you also add some documentation on how to use your class? (If it's too much of a hassle, I can do it as well, no problem).
> 3. Very minor thing: In this library, we generally use CamelCase for types and classes...so could you rename "refl_type" into something like "ReflType"?

check, check and check

liuzicheng1987:
@BjornSAIM , this looks great. 

If the tests run through fine, I will merge it. I'm actually currently on a train and have almost reached my location, so I won't be able to merge it right now. But it will find some time some time today.

Thank you so much for your contribution!

liuzicheng1987:
@BjornSAIM and @DaedieCool , sorry, I had a very busy weekend so I didn't get to merging this until now. But, again, thank you so much for this great addition to the library.

PR #143, opened 2024-07-06
Fix several warnings

BjornSAIM:
Noticed while using Clang 18.1.7


liuzicheng1987:
Looks good to me.

If the checks run through, I will merge this as well.

Issue #150, opened 2024-07-16
Reflector - separate into read and write?


m-mcgowan:
Hi there!

I've been using reflect-cpp for a few weeks, and it's really helping me avoid boilerplate code.  I have an application model, which I'm using reflection to implement a simple CLI where the model can be viewed container hierarchy (with names and types) and values can be retrieved or set. (The CLI works on the desktop and also targeted to embedded devices.)

The recent addition of Reflector support makes customizing serialization/deserialization even easier. However, I have a case where not all types are writable, and it makes no sense to write to them. For example, I have a queue of events, which can be inspected, but not set. (And the queue is not const because parts of the app need to write to the queue.)

One possible way of implementing this is to separate the Reflector into two distinct capabilities - one for reading one for writing. At present, the reflector concept covers both inseparably.

I'd be happy to help submit a PR if these ideas seem like a good way to go.

liuzicheng1987:
Hi @m-mcgowan ,

yeah, I agree, this makes a lot of sense and it is also very easy to do.

All you have to do is to split up the `has_reflector` concept in here into two separate ones (if you could also put them in a separate file `internal/has_reflector.hpp`, that would be great):

https://github.com/getml/reflect-cpp/blob/main/include/rfl/internal/has_reflection_method_v.hpp

Then you have to insert them in the appropriate places in here.

https://github.com/getml/reflect-cpp/blob/main/include/rfl/parsing/Parser_default.hpp

That's it. A PR is welcome, otherwise I can just do it myself.

liuzicheng1987:
And when you do this, if you could also adapt the documentation that would be great:

https://github.com/getml/reflect-cpp/blob/main/docs/custom_parser.md

m-mcgowan:
Thanks for the pointers, I'll give it a shot.

liuzicheng1987:
Resolved.

PR #152, opened 2024-07-17
feat: optional to/from methods on Reflector. (#150)

m-mcgowan:
Changes:

- Separates the `has_reflector` concept into `has_read_reflector` and `has_write_reflector` to allow allow Reflector types to be just readable, just writable, or both readable and writable.

- Adds test cases. The tests are slightly altered from the original `test_reflector` implementation. They now use a small class hierarchy (`Person` and `Parent`) with fields in each class, which is not supported by default. The Reflector implementation in the tests flattens the hierarchy.  Test cases for Reflector implementations with only the `to` and only the `from` method are present.

- Adds a paragraph to the docs mentioning the user-facing changes.

Fixes issue: #150

### DIFF
--- a/docs/custom_parser.md
+++ b/docs/custom_parser.md
@@ -19,8 +19,11 @@ struct Reflector<Person> {
     return {v.first_name, v.last_name};
   }
 };
-} 
+}
 ```
+
+It's also fine to define just the `from` method when the original class is only written, or `to` when the original class is only read.
+
 
 Alternatively, you can implement a custom parser using `rfl::parsing::CustomParser`.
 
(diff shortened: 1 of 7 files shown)


liuzicheng1987:
Looks good to me. I will have to wait for the code checks to run through, but just by looking at the code changes, it looks fine. Thank you so much for implementing this so quickly.

Just one minor thing: In your write_and_read.hpp, you still have this `#if 0 ... #endif` section, which appears to be something you meant to remove, but then forgot. Could you just quickly remove this?

m-mcgowan:
Thanks for catching that.  I wanted to implement `write_and_read` in terms of `read` and `write`, but after a few minutes I gave up :-) 

Unused code removed.

liuzicheng1987:
All right, merged. Thank you for your contribution!

PR #174, opened 2024-08-11
Update README.md to reflect expected libbson version

RyanDraves:
The `libbson` link the README pointed to was the archived version of `libbson`. After struggling to compile things for a while, I realized `reflect-cpp` expects a recent enough version of `libbson` that it needs to originate from https://github.com/mongodb/mongo-c-driver.


liuzicheng1987:
Good catch. Thanks for updating the README.

PR #202, opened 2024-09-29
Split reflect.cpp file into multiple source files

Altren:
Initially I wanted to go deeper with moving more stuff into cpp files where possible, but ClangBuildAnalyzer did not revealed any obvious parts, so I decided to commit only preparation changes.
Any suggestions what can be moved into cpp are welcome, and I'll create another PR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,49 +40,45 @@ project(reflectcpp LANGUAGES CXX)
 
 set(CMAKE_CXX_STANDARD 20)
 
-if(REFLECTCPP_USE_BUNDLED_DEPENDENCIES)
-    if (REFLECTCPP_BUILD_SHARED)
-        add_library(reflectcpp SHARED)
-    else()
-        add_library(reflectcpp STATIC)
-    endif()
+if (REFLECTCPP_BUILD_SHARED)
+    add_library(reflectcpp SHARED)
+else()
+    add_library(reflectcpp STATIC)
+endif()
 
-    target_sources(reflectcpp PRIVATE src/reflectcpp.cpp src/yyjson.c)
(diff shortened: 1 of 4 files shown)


liuzicheng1987:
Hi @Altren , why do you want to split the .cpp file into three smaller files? 

The reason I specifically didn't do this is that I want to make installation as simple as possible. Some people don't use CMake packages, but instead they want to simply include the code in their own build process. 

Being able to simply copy a single source file into your own source code makes this very easy.

liuzicheng1987:
The changes to the CMake file seem fine, but I haven't taken a closer look yet...reducing duplicate code is always a good idea...

Altren:
I splitted this into files, because I wanted to move more code into .cpp files and adding it into single file doesn't looks ok for me.
> Some people don't use CMake packages, but instead they want to simply include the code in their own build process.

Well, they already copy multiple headers anyway. I you want to keep it simple for such use cases I can add reflectcpp.cpp back and `#include` all new .cpp files into it.

liuzicheng1987:
@Altren , at the moment reflectcpp.cpp is fairly short...I don't think there is much of a need to split it up further. It's just going to increase compile time and make things unnecessarily complicated.

However, if you can find significant portions of the code that you can put into source files, we could seriously think about splitting it up. I would be very happy if that were possible.

Altren:
I updated my changes, so that both problem you mentioned are solved:
Compile time is the same as it was, since reflectcpp.cpp is a simple inclusion of other source files.
Only reflectcpp.cpp need to be added to the build, like it was before

Altren:
But I'm ok if we keep this not merged until I'll do more useful changes, that require this split

PR #204, opened 2024-10-03
Move some heavy templates into cpp

Altren:
This PR based on the #202 changes.
Cuts tests build time by 10% on my machine, no hacks or workarounds, just some simple moves to .cpp


liuzicheng1987:
Sounds great!

liuzicheng1987:
@Altren , I think this is a great initiative that makes a lot of sense.

I think a couple of minor changes are still necessary, though:

If we are going to place the JSON code into a source file, we should do the same for all supported formats.

Moreover, we should then determine in the CMakeLists.txt whether we want to compile the source file for a particular format. JSON support is activated by default, but it can be turned off, so we mustn't assume that JSON can be installed.

So I think there should be a source file called reflectcpp_json.cpp, reflectcpp_yaml.cpp, etc

And if the flag for the format is set, the source file will be compiled.

Do you understand what I mean?

Altren:
> If we are going to place the JSON code into a source file, we should do the same for all supported formats.

I agree, I wanted to suggest this change as a next step, since this require more complex changes. But I'm fine doing that in the current PR and will do that.

>Moreover, we should then determine in the CMakeLists.txt whether we want to compile the source file for a particular format. JSON support is activated by default, but it can be turned off, so we mustn't assume that JSON can be installed.
So I think there should be a source file called reflectcpp_json.cpp, reflectcpp_yaml.cpp, etc

This would break the idea of having single source file to compile. But I guess we can keep it simple like:
"If you want to compile reflectcpp with yaml support add reflectcpp.cpp and reflectcpp_yaml.cpp into your build"

I'll do some initial implementation that looks good to me, than we can discuss if it have any disadvantages

Altren:
I've added json.cpp implementation. Please review and if everything looks ok for you I'll continue with other formats

Altren:
Also I feel like reflectcpp_json.cpp and reflectcpp.cpp should not be in the same directory where other source files are, but don't know how to organize this

liuzicheng1987:
@Altren , looks really great. 

One minor thing:

```
if (REFLECTCPP_JSON)
    list(APPEND REFLECT_CPP_SOURCES
        src/reflectcpp_json.cpp
    )
    if (REFLECTCPP_USE_BUNDLED_DEPENDENCIES)
       ....
    else ()
        find_package(ctre CONFIG REQUIRED)
        ...
    endif ()
endif ()
```

CTRE is required regardless of whether we need JSON. So this should be outside the condition `if (REFLECTCPP_JSON)`:

```
  if (REFLECTCPP_USE_BUNDLED_DEPENDENCIES)
      target_include_directories(
          reflectcpp PUBLIC
          $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/rfl/thirdparty>)
  else ()
      find_package(ctre CONFIG REQUIRED)
  endif ()

if (REFLECTCPP_JSON)
   ...
endif()
```

Altren:
There are only two new commits above (XML to cpp) and squashed changes about ctre

Altren:
Should I continue with other formats in this PR or better do them in a separate on to avoid too big PR?

Altren:
^ rebased onto main to resolve conflicts

Altren:
Just in case, now this is totally complete, unless there are more review comments

liuzicheng1987:
@Altren , merged. Thank you for your contribution.

PR #211, opened 2024-10-08
add simple precompiled headers

Altren:
Cuts tests and benchmarks build time by ~30%
Does not affect library users


liuzicheng1987:
Merged. Thanks so much.

PR #219, opened 2024-10-15
Fix endless recursion in std::string_view writer

dmikushin:
Fix endless recursion in `std::string_view` writer: the `write()` should use `std::string`, not `std::string_view` again


liuzicheng1987:
Good catch...could you quickly add a test for that? I always thought we had a test for string_views. (JSON only is fine.)

Issue #224, opened 2024-10-21
Wrong filename in logic checking for local `ctre.hpp`


Harry-Chen:
Should check for `ctre.hpp`, not `ctre.h` in:

https://github.com/getml/reflect-cpp/blob/b3b0370ce2737336d0482cf6a23d4eb7daf75a3e/include/rfl/PatternValidator.hpp#L6-L10

liuzicheng1987:
Great catch. Feel free to open a PR, otherwise I will fix it tonight.

PR #225, opened 2024-10-21
Fix wrong header filename for ctre library

Harry-Chen:
Fixes getml/reflect-cpp#224


liuzicheng1987:
@Harry-Chen thank you so much for the fix!

PR #232, opened 2024-10-26
Increased maximum number of fields from 100 to 256.

### LINKED ISSUE #154: Bug: number of fields in a struct is limited to 100; consider removing limitation?
I'm not certain if this is a limitation in the library itself or in compilers, but I do have a use case where I've exceeded this limit significantly and I don't really have a way of splitting the fields up since Idk what I'd even name the substructs. 😀 If this is a limitation at the compiler level, I believe it can be raised for most of them:

* MSVC: fixed at 2048
* Clang, GCC, Intel: `-ftemplate-depth=<N>`

If this is just a library-imposed limit, may I ask why it is there?

danielmohansahu:
Closes #154 by increasing the number of supported fields.


liuzicheng1987:
@danielmohansahu , thanks for your initiative and your contribution!

PR #246, opened 2024-11-04
Automatically add tags to variants; fixes #230

### LINKED ISSUE #230: std::variant serialization of empty types can be irreversible
the variant's current alternative type is not serialized, making it impossible to recover when deserializing
Making this sort of use case where tagged unions are sometimes left empty unusable.

There are however no compile errors, so i believe this should either be supported or throw a compiler error.

```cpp
    struct button_pressed_t
    {
        bool operator==(const button_pressed_t&) const = default;
    };

    struct button_released_t
    {
        bool operator==(const button_released_t&) const = default;
    };

    struct key_pressed_t
    {
        char key;

        bool operator==(const key_pressed_t&) const = default;
    };

    using my_event_type_t = std::variant<button_pressed_t, button_released_t, key_pressed_t>;
    struct my_event_t
    {
        my_event_type_t event;

        [[nodiscard]] constexpr bool operator==(const my_event_t& other) const noexcept = default;
    };

    auto my_event = my_event_t{.event = key_pressed_t{}};

    auto as_json = rfl::json::write(my_event, rfl::json::pretty);
    auto reparsed_event = rfl::json::read<my_event_t>(as_json).value();

    // Should be equal
    ASSERT_NE(my_event, reparsed_event);
```

liuzicheng1987:


docs/processors.md
```diff
@@ -61,6 +62,58 @@ The resulting JSON string looks like this:
 {"type":"Person","first_name":"Homer","last_name":"Simpson","age":45}
 ```
 
+### `rfl::AddStructsToVariants`
```
bryceschober, line 65: Typo: `AddStructsToVariants` -> `AddTagsToVariants`

docs/processors.md
```diff
@@ -61,6 +62,58 @@ The resulting JSON string looks like this:
 {"type":"Person","first_name":"Homer","last_name":"Simpson","age":45}
 ```
 
+### `rfl::AddStructsToVariants` 
+
+This processor automatically adds structs to variants. Consider the following example:
```
bryceschober, line 67: Suggestion: "automatically adds struct tags to"

docs/processors.md
liuzicheng1987: Thanks!

docs/processors.md
liuzicheng1987: That was meant to be "tags" as well...

PR #247, opened 2024-11-05
Fixed issue where 32 bit compilation cause template type deduction fail

dcorbeil:
There was a compilation error when building on 32 bit platforms. I assume it is because `std::size_t` used for `std::array<T, std::size_t N>` is an `unsigned long` on 64 bit and `unsigned int` on 32 bit. 

See this for a quick demonstration: https://godbolt.org/z/xcrdjM5Ks


liuzicheng1987:
Thanks for your contribution!

PR #261, opened 2024-11-15
Added new mkdocs based documentation to the project

alxn4:


liuzicheng1987:
@alxn4 , looks great, but there are a few minor things to be fixed:

- The link to the benchmarks in the Design Principles section (both README and Documentation) leads to a 404.
- The headers in the README are messed up: "Feature Overview" should be an L2 header and all headers below that L3, but it is an L1 header.
- The three links in the second paragraph of the Installation section of the documentation don't seem to be doing anything.
- "reflect-cpp in a nutshell" seems to be a pointless header, because it consists of only one subheader "Why do we need this?" (which more accurately summarizes the text)
- In the last paragraph in section on structs of the documentation, there are two links to custom classes and custom parsers - they both point back to the Github repo.

PR #275, opened 2024-11-23
Fix the msgpack dependency as well

liuzicheng1987:


CMakeLists.txt
```diff
@@ -135,7 +135,11 @@ if (REFLECTCPP_MSGPACK)
     list(APPEND REFLECT_CPP_SOURCES
         src/reflectcpp_msgpack.cpp
     )
-    find_package(msgpack-c CONFIG REQUIRED)
+    if(REFLECTCPP_USE_VCPKG)
+        find_package(msgpack-c CONFIG REQUIRED)
+    else()
+        find_package(msgpack CONFIG REQUIRED)
```
mirageAlchemy, line 141: Hi @liuzicheng1987 , I am not quite familar with the context, could you be kind and explain to me why we have to have an if branch here? I bumped into this issue when I try to use msgpack support with xmake and I have to fix it by manually applying a diff on the CMakeLists.txt: https://github.com/xmake-io/xmake-repo/pull/6205/files. Thank you.

liuzicheng1987: @mirageAlchemy , the issue is tha in vcpkg you have to find "msgpack-c", but in Conan you have to find "msgpack".

https://conan.io/center/recipes/msgpack-c?version=6.1.0

Apparently in XMake it is called msgpack-c as well...what a mess...

mirageAlchemy: Ahh I see, personally I think Conan choose a wrong name, because it also hosts `msgpack-cxx`too. Probably because previously there is only [msgpack](https://conan.io/center/recipes/msgpack) on conan, but later when someone wants to add the cpp implementation, it was split into `msgpack-c` and `msgpack-cxx`, and the name was kept intact to maintain backward compatibility.

mirageAlchemy: Do you think it is a good idea to add a config called `msgpack_target_name` and default to `msgpack` unless REFLECTCPP_USE_VCPKG is set? It would make later update much easier for me on the XMake side.

PR #296, opened 2024-12-14
Timestamp: Update workaround to support MinGW alongside MSVC

Doekin:
Original
---
Timestamp: Replace `strptime` with `std::get_time` for better compatibility

The `strptime` function is a POSIX extension, not universally supported by all compilers (e.g., MSVC). Replacing it with `std::get_time`, which is part of the C++ standard, improves cross-platform compatibility without sacrificing functionality.

Edit
---

`std::get_time` has poor performance compared to `strptime`. My issue was that MinGW failed to compile `reflect-cpp` due to the lack of `strptime`. A more precise solution is to extend the existing workaround to cover MinGW, while keeping `strptime` for better performance on supported platforms.

### DIFF
--- a/include/rfl/Timestamp.hpp
+++ b/include/rfl/Timestamp.hpp
@@ -76,20 +76,6 @@ class Timestamp {
   const std::tm& tm() const { return tm_; }
 
  private:
-#ifdef _MSC_VER
-  // This workaround is necessary, because MSVC doesn't support strptime.
-  char* strptime(const char* _s, const char* _f, std::tm* _tm) {
-    std::istringstream input(_s);
-    input.imbue(std::locale(setlocale(LC_ALL, nullptr)));
-    input >> std::get_time(_tm, _f);
-    if (input.fail()) {
-      return NULL;
-    }
-    return (char*)(_s + input.tellg());
-  }
-#endif
-
- private:
   /// The underlying time stamp.
   std::tm tm_;
 };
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
Hi, my concern with the get_time solution is that strptime is more efficient. I'd rather only use strptime where possible and use the workaround where necessary.

Or is there any specific problem you encountered because of that?

Doekin:
Hi,  
I encountered a compilation issue while using `reflect-cpp` with MinGW:  
```
error: use of undeclared identifier 'strptime'
```

Since `strptime` is a POSIX extension, it isn’t universally supported and may require additional workarounds for certain platforms. Using a standard function like `std::get_time` would improve cross-platform compatibility out of the box.

That said, I understand your concern about efficiency. I’ll run a benchmark to evaluate the performance difference and see if the trade-off is reasonable. Let me follow up with the results soon.

Doekin:
[A quick benchmark](https://quick-bench.com/q/bAsSWchhVIXtfRBM3SvistYaRmw) shows `std::get_time` is quite slower than `strptime`. Sorry, I should’ve looked into this performance difference earlier.

liuzicheng1987:
@Doekin, no problem. Let's try to resolve your issue by finding a macro that captures your case as well.

Doekin:
`x86_64-w64-mingw32-gcc` defines both `__MINGW32__` and `__MINGW64__`. It seems that using `__MINGW32__` might be sufficient to capture this case.

liuzicheng1987:
@Doekin, merged. Thanks for your contribution!

PR #298, opened 2024-12-17
Use static_cast for the pointer transformations

liuzicheng1987:


GregTheMadMonk:
That should be legal... I guess?

https://eel.is/c++draft/conv.ptr#2
```
A prvalue of type “pointer to cv T”, where T is an object type, can be converted to a prvalue of type “pointer to cv void”[.](https://eel.is/c++draft/conv.ptr#2.sentence-1) The pointer value ([[basic.compound]](https://eel.is/c++draft/basic.compound)) is unchanged by this conversion[.](https://eel.is/c++draft/conv.ptr#2.sentence-2)
```

https://eel.is/c++draft/expr.static.cast#13
```
A prvalue of type “pointer to cv1 void” can be converted to a prvalue of type “pointer to cv2 T”, where T is an object type and cv2 is the same cv-qualification as, or greater cv-qualification than, cv1[.](https://eel.is/c++draft/expr.static.cast#13.sentence-1) If the original pointer value represents the address A of a byte in memory and A does not satisfy the alignment requirement of T, then the resulting pointer value ([[basic.compound]](https://eel.is/c++draft/basic.compound)) is unspecified[.](https://eel.is/c++draft/expr.static.cast#13.sentence-2) Otherwise, if the original pointer value points to an object a, and there is an object b of type similar to T that is [pointer-interconvertible](https://eel.is/c++draft/basic.compound#def:pointer-interconvertible) with a, the result is a pointer to b[.](https://eel.is/c++draft/expr.static.cast#13.sentence-3) Otherwise, the pointer value is unchanged by the conversion[.](https://eel.is/c++draft/expr.static.cast#13.sentence-4)
[Example [3](https://eel.is/c++draft/expr.static.cast#example-3): 
T* p1 = new T;
const T* p2 = static_cast<const T*>(static_cast<void*>(p1));
bool b = p1 == p2;  // b will have the value true.
— end example]
```

As long as the actual stored data at the pointer is of correct type...

liuzicheng1987:
@GregTheMadMonk, thanks for your feedback. Very interesting!

It certainly works in practice and seems to be fine in theory. I'm going to merge this.

GregTheMadMonk:
Hopefully I didn't mislead us both xD

ZXShady:
I find it weird since the double static cast is equalivent to a reinterpret_cast

PR #312, opened 2024-12-22
Json read string view

aligirayhanozbay:
Hi, this is a very simple PR to change rfl::json::read to use std::string_view instead of std::string. This allows JSON to be read in a zero-copy way when e.g. one has a const char*. Because std::string_view can be implicitly constructed from std::string, the change should have no impact on existing user code.

I verified that all json parser tests pass with this change.

### DIFF
--- a/include/rfl/json/read.hpp
+++ b/include/rfl/json/read.hpp
@@ -8,7 +8,6 @@
 #endif
 
 #include <istream>
-#include <string>
 #include <string_view>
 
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
@aligirayhanozbay, looks great. I will run the test pipelines and if they all pass, I will merge this.

liuzicheng1987:
@aligirayhanozbay, thanks for the contribution!

PR #319, opened 2025-01-02
Update methods of getting dependencies

xbrtszstk:
Hello,
this time I updated methods of getting dependencies.
These methods work with Yocto build system, where I also added some recipes for dependencies used by this project.
You can take a look on PR for meta-oe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,13 +106,18 @@ endif ()
 
 if (REFLECTCPP_AVRO)
     list(APPEND REFLECT_CPP_SOURCES
-        src/reflectcpp_avro.cpp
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
Hi @xbrtszstk , I don't this can work...the tomlplusplus recipe does not provide a cmake.config on vcpkg.io (but it does on Conan). 

As far as TinyCBOR is concerned, I am probably just going to use jsoncons instead, just to get around this headache.

liuzicheng1987:
Maybe the best approach for now is to support only some of the formats. Maybe we can figure out a way to deal with Avro and CBOR some other time. It's not like they are terribly important formats anyway.

PR #333, opened 2025-01-13
`rfl::Generic`: add support for long (int64)

bjia56:
While working on code that converts `rfl::Generic` to JSON and back, I found that `reflect-cpp@v0.17.0` does not properly convert int64's through Generic. Namely, Generic is defined as a variant with `int` as one option, which causes the JSON parser to cast yyjson values to `int`, potentially truncating the number.

This change introduces `long` as a variant option under Generic. It is placed before `int` to prioritize converting numbers as the bigger variant in the JSON parser, as otherwise the parser will indiscriminately match all integers to int. However, this likely has the side effect where all ints are now treated as longs. As a corrective measure, `to_int` and `to_long` will fall back and static cast from the other type if the variant does not directly match. I was usure of the impact of upgrading the variant's int to long (instead of the proposed implementation of adding a new variant), but perhaps replacing int with long would be cleaner?

If an alternative implementation would be better, please let me know!

### DIFF
--- a/include/rfl/Generic.hpp
+++ b/include/rfl/Generic.hpp
@@ -19,10 +19,10 @@ class Generic {
 
   using Array = std::vector<Generic>;
   using Object = rfl::Object<Generic>;
-  using VariantType = std::variant<bool, int, double, std::string, Object,
+  using VariantType = std::variant<bool, int, long, double, std::string, Object,
                                    Array, std::nullopt_t>;
   using ReflectionType = std::optional<
-      std::variant<bool, int, double, std::string, Object, Array>>;
+      std::variant<bool, int, long, double, std::string, Object, Array>>;
 
   Generic();
 
@@ -94,6 +94,10 @@ class Generic {
   /// underlying value is not an integer.
   Result<int> to_int() const noexcept;
 
+  /// Casts the underlying value to a long or returns an rfl::Error, if the
+  /// underlying value is not a long.
+  Result<long> to_long() const noexcept;
+
   /// Casts the underlying value to an rfl::Generic::Object or returns an
   /// rfl::Error, if the underlying value is not an rfl::Generic::Object.
   Result<Object> to_object() const noexcept;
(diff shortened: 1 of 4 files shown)


liuzicheng1987:
@bjia56 , some of the tests fail, could you take a look?

smurfomen:
I have same problem too.
When using rfl::msgpack for serialization and deserialization with a Variant type, numeric types (e.g., uint64_t, int64_t) are being deserialized as double instead of their original types. This results in a loss of type information and incorrect representation of the data.

```cpp
using Var = rfl::Variant<double, uint64_t, int64_t, bool, std::string>;
struct Item
{
int64_t id;
Var data;
};

Item i {.id = 2, .data = static_cast<int64>(42) };
```
And I take 
```json
{
  "id": 2,
  "data": 42.0
}
```

If I change the sequence of types (like int64_t, uint64_t, double, others ), then on the contrary I will always get int64_t.

liuzicheng1987:
@smurfomen , this is a different problem. The problem here is that the variant tries the different types one by one. And since a double can be cast to an int64_t and vice versa, it will do that.

You could try std::variant instead, but I am not sure it would behave differently.

liuzicheng1987:
@smurfomen , could you try this?

```cpp
template <class T>
struct Tag {
    using ReflectionType = T;
    T value;
};

using Var = rfl::Variant<Tag<double>, Tag<uint64_t>, Tag<int64_t>, Tag<bool>, Tag<std::string>>;

struct Item
{
int64_t id;
Var data;
};

Item i {.id = 2, .data = Tag<int64>(42) };
```

bjia56:
I'm coming back to this to take a look at the failures. In my own uses of reflect-cpp, I've actually switched Generic to use int64_t as a wholesale replacement for int, which works better on Windows (since MSVC defines long as 32-bit).

bjia56:
@liuzicheng1987 The tests now seem to pass when run on my Ubuntu setup. Can you please try the GH actions runs again?

liuzicheng1987:
@bjia56 , looks great. Thank you so much for your contribution.

PR #348, opened 2025-02-12
cmake: unify library name of msgpack-c, use pkgconf to find libavro

Harry-Chen:
* msgpack uses cmake module name `msgpack` with conan, and `msgpack-c` with vcpkg and native package (e.g. libmsgpack-c-dev in Debian)
* avro provides a pkgconfig `avro-c.pc`


liuzicheng1987:
@Harry-Chen , thank you very much!

PR #362, opened 2025-02-24
Parser_default: Move temporary variable instead of copying

hlewin:
This eliminates an unnecessary copy of a temporary variable which prohibits use of a move-only construction of reflection-types.


liuzicheng1987:
Thanks!

PR #369, opened 2025-02-28
fix build with llvm-18 by changing Bytestring type to std::vector instead of deprecated std::basic_string<std::byte>

Altren:


liuzicheng1987:
@Altren , thanks for the contribution. 

I understood the problem. Could you also update the documentation?

liuzicheng1987:
@Altren thank you so much!

PR #373, opened 2025-03-07
add missing include <new>

filippobrizzi:
I found this issue while compiling with clang 20.1.0.
Header <new> is needed by `std::launder`


liuzicheng1987:
Thanks for the contribution!

PR #374, opened 2025-03-07
add optional yyjson_read_flag symmetric to the yyjson_write_flag, so that non standard json's can be both written and read

Altren:


liuzicheng1987:
Makes sense

PR #390, opened 2025-03-22
Fix bugging disallowing optionals to hold items without copy constructors

jmcken8:
Also adds a test.


liuzicheng1987:
@jmcken8 Thanks for your contribution!

PR #392, opened 2025-03-23
Add enum range support with customizable min and max values

jokla:
Currently, the supported value range in reflect-cpp is 0 to 127. This change aims to allow users to customize the enum range.

Enum values must be in the range `[RFT_ENUM_RANGE_MIN, RFT_ENUM_RANGE_MAX]`. If the range is not specified, the
   default range is `[0, 127]`.

- You can specify a custom range for the all enum values by defining `RFT_ENUM_RANGE_MIN` and `RFT_ENUM_RANGE_MAX`
      before including the reflect-cpp header:

        #define RFL_ENUM_RANGE_MIN -128
        #define RFL_ENUM_RANGE_MAX 128
        #include <rfl.hpp>

- You can specify a custom range for a specific enum by defining the specialization `rfl::config::enum_range` for
      the enum type :

      enum class Color { yellow = 200, purple = 300, orange = 400 };
   
      template <>
      struct rfl::config::enum_range<Color> {
        static constexpr int min = 200;
        static constexpr int max = 400;
      };

### DIFF
--- a/docs/enums.md
+++ b/docs/enums.md
@@ -27,34 +27,42 @@ This results in the following JSON string:
 
 However, some limitations apply:
 
-1. They must be be scoped enumerations.
+1. They must be scoped enumerations.
 
-```cpp
-/// OK - scoped enumeration
-enum class Color1 { red, green, blue, yellow };
+    ```cpp
+    /// OK - scoped enumeration
+    enum class Color1 { red, green, blue, yellow };
+    
+    /// OK - scoped enumeration
+    enum struct Color2 { red, green, blue, yellow };
+    
+    /// unsupported - unscoped enumerations
+    enum Color3 { red, green, blue, yellow };
+    ```
 
-/// OK - scoped enumeration
-enum struct Color2 { red, green, blue, yellow };
+2. Enum values must be in the range `[RFL_ENUM_RANGE_MIN, RFL_ENUM_RANGE_MAX]`. If the range is not specified, the
+   default range is `[0, 127]`.
 
-/// unsupported - unscoped enumerations
-enum Color3 { red, green, blue, yellow };
-```
+    - You can specify a custom range for the all enum values by defining `RFL_ENUM_RANGE_MIN` and `RFL_ENUM_RANGE_MAX`
+      before including the reflect-cpp header:
 
-2. You cannot have more than 128 values and if you explicitly assign values, they must be between 0 and 127.
-
-```cpp
-/// OK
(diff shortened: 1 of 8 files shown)


liuzicheng1987:
Hi @jokla,

This looks great, thank you so much.

My only comment is that we should always test one thlng for every test, so if you could break up your test into three different ones, that would be great.

jokla:
Hi @liuzicheng1987, thank you for your feedback! 

I can definitely change the tests. I just have one question, they are already divided into three different ones. Would you prefer that I create separate files for them?

liuzicheng1987:
@jokla , yes, please divide them into separate files, that is what I meant.

By the way, this morning I was working on a way to remove the recursion from the function that infers the names. I think I will get it done by tomorrow. That might gives a bit more leeway.

jokla:
Awesome @liuzicheng1987 I will separate the tests into different files and wait for the new implementation!

liuzicheng1987:
@jokla it's in PR #395. 

There is something weird going on with Conan on macOS, but I am pretty sure it's unrelated to the changes I made.

Feel free to pull the branch over into yours.

liuzicheng1987:
@jokla the fold expressions are merged into main. Feel free to merge them over into your branch. You're likely going to have a merge conflict, but it shouldn't be too hard to resolve. After that, we can merge this as well.

jokla:
It looks like the current implementation has brought up another limit issue with some compilers. I didn’t run into this on my local machine with gcc-13. 

> Error: /home/runner/work/reflect-cpp/reflect-cpp/include/rfl/internal/enums/get_enum_names.hpp:149:28: fatal error: instantiating fold expression with 257 arguments exceeded expression nesting limit of 256

What do you suggest to do? I am trying something out here [4e48e1f](https://github.com/getml/reflect-cpp/pull/392/commits/4e48e1f4a16de47f42d93bc2f8f0154ae42445b5) but I am not sure if it is an acceptable solution.  It passed all the checks except for Windows. I am not sure if this is related to the change; I will have to look into it later.

Many thanks!

liuzicheng1987:
@jokla, unfortunately, the GitHub Windows runner is sometimes a bit unstable and likes to time out. In these cases, just re-run it.

liuzicheng1987:
@jokla I re-ran the Windows pipeline and it passed. 

Great work, absolutely love it!

I think this is ready to be merged. Any objections?

jokla:
Great @liuzicheng1987 ! I am happy to have this merged!

Issue #397, opened 2025-03-25
Using 'Box' makes data non-copyable


jmcken8:
An issue I've run into a few times with this library is that if/when you need to use `rfl::Box` (i.e for recursion), it makes copying any data structure containing a box (at any nesting level) impossible without defining copy constructors for the class/struct containing the `rfl::Box`, which can be difficult to get right so that this library still functions.

Being able to trivially copy data is often useful.

I could see two ways to implement this:

1. Make an explicitly copyable box implementation
2. Use the [copy constructable](https://en.cppreference.com/w/cpp/concepts/copy_constructible) concept to conditionally enable copy constructors for the rfl::Box

The second option is definitely 'easier' for use but would change the behaviour of Box pretty extensively and may break existing code, is this project resistant to breaking changes of this sort?

jmcken8:
I'm happy to post a MR for the second option as well

jmcken8:
Hmm actually, thinking about it, I don't think it's possible to do 2. Whatever you do, if you try to rely on detecting whether a type is copy-constructible, with a recursive type you'll always run into the fact that you have an incomplete class definition when you get back to the original type, which means for copying to work you must explicitly declare either the box or the class itself copyable.

liuzicheng1987:
@jmcken8 , the fact that `rfl::Box` cannot be copied is by design. 

It wraps a `std::unique_ptr` and therefore is supposed to have unique ownership of the underlying object. If it were copyable, you would run into double-frees.

And if copying `rfl::Box` also copied the underlying object, why would you want it in the first place?

I think what you are looking for is `rfl::Ref`.

jmcken8:
> And if copying rfl::Box also copied the underlying object, why would you want it in the first place?

Box is required for defining safe recursive structures, which is the main stated use of `rfl::Box` in the documentation. If it's just supposed to be a "non-null std::unique_ptr" then I agree with your points, but there are plenty of cases where it's useful for a recursive structure to be copyable. Since the project doesn't really support other uses of `std::unique_ptr` like polymorphic types, pImpl idiom, etc., and any structure that's self-referential can't be output to JSON, I would think in nearly all cases structs reflectable by `reflect-cpp` _should_ be copyable, with currently recursion being the main exception.

For an example from the documentation at https://rfl.getml.com/rfl_ref/, if you wanted to copy and update a 'Person' struct, you can just do:

```cpp
struct Person {
    rfl::Rename<"firstName", std::string> first_name;
[...]

PR #398, opened 2025-03-25
Possible implementation for #397: Allow Boxes to conditionally be copyable with an explicit option

jmcken8:
Possible implementation for: https://github.com/getml/reflect-cpp/issues/397

With an explicit type `rfl::CopyableBox` that can be substituted in for `rfl::Box`, that enables easy copying of data.

### DIFF
--- a/docs/rfl_ref.md
+++ b/docs/rfl_ref.md
@@ -1,4 +1,4 @@
-# `rfl::Box` and `rfl::Ref` 
+# `rfl::Box` and `rfl::Ref`
 
 In previous sections, we have defined the `Person` class recursively:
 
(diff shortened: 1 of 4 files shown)


jmcken8:
Running clang-format changes the files pretty extensively so I avoided auto-formatting for this MR

jmcken8:
Docs would need to be updated if this MR is accepted, happy to do so

liuzicheng1987:
But what is the point of this? If you want this, why not just use `rfl::Ref`?

liuzicheng1987:
Cool, thank you so much for your contribution.

PR #401, opened 2025-04-08
Change back to tomlplusplus

liuzicheng1987:
For a long time, tomlplusplus did support a config.cmake file, which made it difficult to use on vcpkg. As a fix, we switched to toml11.

But this has now finally been added to vcpkg, and since tomlplusplus is superior to toml11 in terms of compile time and runtime, we are changing back.


liuzicheng1987:
For a long time, tomlplusplus did generate a config.cmake file, which made it difficult to use on vcpkg. As a fix, we switched to toml11.

But this has now finally been added to vcpkg, and since tomlplusplus is faster than toml11 in terms of compile time and runtime, we are changing back.

PR #403, opened 2025-04-09
fix bytestring being serialized as vector of flag enums

Altren:
This is a fix for the regression introduced with the Bytestring std::vector replacement.
It was serialized as a vector of enums instead of binary data.
I haven't found a cleaner way other than making a fake type to avoid it being treated as `std::vector<T>`.
Also, I can make write_and_read return serialized size, so that such regression can be caught by the test_bytestring and explicit size check in it

### DIFF
--- a/include/rfl/Bytestring.hpp
+++ b/include/rfl/Bytestring.hpp
@@ -5,8 +5,13 @@
 #include <vector>
 
 namespace rfl {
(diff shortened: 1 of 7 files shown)


liuzicheng1987:
@Altren , thank you for your contribution!

PR #425, opened 2025-05-02
Allow passing yyjson flags in the read(istream) method

AndrewNolte:


liuzicheng1987:
@AndrewNolte great catch

liuzicheng1987:
@AndrewNolte thanks for the contribution

PR #431, opened 2025-05-06
Upgrade yyjson to 0.11.0

BestITUserEUW:


liuzicheng1987:
@BestITUserEUW could you also update the vcpkg.json and Conanfile.py?

BestITUserEUW:
@liuzicheng1987 Seems like conan first needs have yyjson in their registry available. Why does ctre and yyjson need to be downloaded if they come bundled with reflect-cpp ?

liuzicheng1987:
@BestITUserEUW

BestITUserEUW:
@liuzicheng1987 ok i will close this pull request here until it is available

liuzicheng1987:
Sorry, I wasn't sufficiently clear. I think you can just go back to YYJSON 0.10.0 in the Conanfile and upgrade everything else.

@BestITUserEUW

PR #441, opened 2025-05-28
Add `enchantum` as an enum reflection backend

ZXShady:
Disclaimer I am the author of [enchantum](github.com/ZXShady/enchantum).

This PR replaces reflect-cpp own enum reflection with enchantum's

1. It reduces compile times
2. It reduces the need to maintain this backend
3. it allows for more customization you can for example choose to remove a prefix on all enums by using `enchantum::enum_traits`
4. it allows reflection of all enums types
(except unnamed unscoped enums) and not just scoped enums.

```cpp
enum {
  Hello // Unreflectable the main enum is unnamed
};
```


include/rfl/internal/enums/Names.hpp
```diff
@@ -1,6 +1,32 @@
 #ifndef RFL_INTERNAL_ENUMS_NAMES_HPP_
 #define RFL_INTERNAL_ENUMS_NAMES_HPP_
 
+// Enum values must be greater than or equal to RFL_ENUM_RANGE_MIN.
+// By default, RFL_ENUM_RANGE_MIN is set to 0.
+// To change the default minimum range for all enum types, redefine the macro
+// RFL_ENUM_RANGE_MIN.
+#if !defined(RFL_ENUM_RANGE_MIN)
+#define RFL_ENUM_RANGE_MIN -256
```
bryceschober: Suggestion: The range `(-256, 256)` strikes me as rather odd, since it doesn't match the previous implementation, and also goes over the one-byte "boundary" of an underlying type. I haven't looked at your `enchantum` internals yet, but I suspect that this new > 9-bit enum range may have some performance implications.

ZXShady: my library clamps the enum ranges so it doesn't matter.


I did a small test with magic enum, mine, and rfl in this gist

https://gist.github.com/ZXShady/74ebfe7d14ddcfeee6b0577a9ee4a396

time to compile

magic enum : 9 seconds

enchantum : 2.5  seconds

rfl : 19 seconds



and by performance you mean what exactly? compile time or runtime?

bryceschober: Both compile time and also memory, but I suppose that would be driven by the underlying type more than the enum range?

ZXShady: > Both compile time and also memory, but I suppose that would be driven by the underlying type more than the enum range?

no the enum range determines the compile times. see my library readme

but if enum range exceeds underlying type min max it clamps the range.

the compile time are faster (see gist above) and memory usage should be the same, memory usage while compiling though is heavily reduced just like compile times

liuzicheng1987:
Unless the tests reveal some compiler/OS-specific problems we haven't discovered so far (unlikely), this is ready to be merged, IMO.

@ZXShady , do you agree?

@ZXShady , I have one more request: When you upload enchantum to vcpkg and Conan, please notify me. The maintainers of these package managers want to avoid us including thirdparty libraries that they also ship. If people have two different versions of the same library in their code and don't even know about it that can cause all sort of problems...

ZXShady:
> Unless the tests reveal some compiler/OS-specific problems we haven't discovered so far (unlikely), this is ready to be merged, IMO.
> 
> @ZXShady , do you agree?


I was thinking about using the split headers instead of the big one since I always keep forgetting to update it and it bloats compile times what do you think?
and we would have to wait for the tests.


> @ZXShady , I have one more request: When you upload enchantum to vcpkg and Conan, please notify me. The maintainers of these package managers want to avoid us including thirdparty libraries that they also ship. If people have two different versions of the same library in their code and don't even know about it that can cause all sort of problems...

I am thinking of writing a vcpkg port.

ZXShady:
Github sctions seems to be failing for me as well right now. been waiting and no runner been picked up

liuzicheng1987:
@ZXShady what exactly does the prefix length do? I think it can be easily added to the configurations...

ZXShady:
@liuzicheng1987 

it removes N characters from the start of enum name.

example in docs

```cpp
#include <enchantum/enchantum.hpp>
#include <enchantum/bitwise_operators.hpp>
// this is a bug enum it is outside of default range [-256,256] and it has this annoying prefix
enum BigEnumOutsideOfDefault : std::uint16_t {
  BigEnumOutsideOfDefault_A =0,BigEnumOutsideOfDefault_B = 4096 
};

template<>
struct enchantum::enum_traits<BigEnumOutsideOfDefault> {
  constexpr static auto prefix_length = sizeof("BigEnumOutsideOfDefault_")-1;
  constexpr static auto min = 0;
  constexpr static auto max = 4096;
};


// prefix removed
static_assert(enchantum::names<BigEnumOutsideOfDefault>[0] == "A");
static_assert(enchantum::names<BigEnumOutsideOfDefault>[1] == "B");
```


I anm myself against making aliases for existing things. if the user needs this he should just specialize enchantum's instead.

liuzicheng1987:
@ZXShady I See. I suppose that is something we can always add later.

liuzicheng1987:
@ZXShady the tests have finally succeeded. I don't why they take do long these days.

But I am merging. Thank you so much for this great contribution. If you really manage to extend the range to -32,000 to 32,000, please open another PR. I would be very interested.

ZXShady:
> @ZXShady the tests have finally succeeded. I don't why they take do long these days.
> 
> But I am merging. Thank you so much for this great contribution. If you really manage to extend the range to -32,000 to 32,000, please open another PR. I would be very interested.

The extended range will just require updating the header files btw.

(I have actually implemented it currently for clang, it is still local though on my machine though)

Issue #444, opened 2025-06-08
Don't use `std::string_view` with APIs expecting null-terminated strings


ZXShady:
```cpp
template <class T, class... Ps>
Result<T> read(const std::string_view _xml_str) {
  pugi::xml_document doc;
  const auto result = doc.load_string(_xml_str.data());
  if (!result) {
    return error("XML string could not be parsed: " +
                 std::string(result.description()));
  }
  const auto var = InputVarType(doc.first_child());
  return read<T, Ps...>(var);
}
```

https://github.com/getml/reflect-cpp/blob/main/include%2Frfl%2Fxml%2Fread.hpp#L32-L42

`doc.load_string` expects a null terminated string so it knows the length, this could easily lead to security vulnerabilities with non-null yerminated inputs

Solutions: 

- Take `const std::string&` as parameter instead. (cheap if user already passes a `std::string` but limits the user from passing pther string types implicitly.

- Keep `std::string_view` in parameter but construct a `std::string` inside the function (more expensive)

- Use another function internally that accepts a `const char* str`,`std::size_t len` pair

liuzicheng1987:
@ZXShady , great catch, thank you so much. Issue is resolved with ec6ab3baf5d42bec0e2f680ddd65116184a8f28f.

PR #445, opened 2025-06-08
Improved support for string_views for TOML and XML

liuzicheng1987:


ZXShady:
my issue was not comprehensive,


there is some other places like

https://github.com/getml/reflect-cpp/blob/main/src%2Frfl%2Fyaml%2FWriter.cpp#L65

liuzicheng1987:
@ZXShady , the writer should be fine. It's not exposed to the end-users. These string views are actually derived from the field names of our structs.

ZXShady:
> @ZXShady , the writer should be fine. It's not exposed to the end-users. These string views are actually derived from the field names of our structs.

shouldn't it be in the internal namespace then?

liuzicheng1987:
@ZXShady , I don't really think so. I actually do encourage people to write their own readers and writers and go at great length to explain how to do so. It's just that no one would ever directly use them. They would be exposed to the library-internal code. It's pretty similar to your custom enum ranges.

Basically, the string_views in the writers are the field names derived from the structs. They originally come in the form of a StringLiteral:

https://github.com/getml/reflect-cpp/blob/main/include/rfl/internal/StringLiteral.hpp

As you can see, they are null-terminated, pretty much precisely for this reason. Where we can use the size, we do, but if we can't, it's not a big problem. I understand this is a potential footgun, but the reality is that yaml-cpp does not support string views. (Let alone any of the C libraries.) If we were to replace them with std::string, this would basically force us to always create a deep copy, which is something I would very much like to avoid.

However, xml::read is exposed to the end user, of course, which is why your issue (https://github.com/getml/reflect-cpp/issues/444) is a really great catch! Thank for this!

ZXShady:
> @ZXShady , I don't really think so. I actually do encourage people to write their own readers and writers and go at great length to explain how to do so. It's just that no one would ever directly use them. They would be exposed to the library-internal code. It's pretty similar to your custom enum ranges.
> 
> Basically, the string_views in the writers are the field names derived from the structs. They originally come in the form of a StringLiteral:
> 
> https://github.com/getml/reflect-cpp/blob/main/include/rfl/internal/StringLiteral.hpp
> 
> As you can see, they are null-terminated, pretty much precisely for this reason. Where we can use the size, we do, but if we can't, it's not a big problem. I understand this is a potential footgun, but the reality is that yaml-cpp does not support string views. (Let alone any of the C libraries.) If we were to replace them with std::string, this would basically force us to always create a deep copy, which is something I would very much like to avoid.

https://github.com/jbeder/yaml-cpp/blob/master/include%2Fyaml-cpp%2Femitter.h#L76

has an overload.
But I see what you mean the strings are the Literals therefore always null terminated.

_When will C++ get zstring_view accepted in the standard library_

liuzicheng1987:
@ZXShady , turns out it actually has a direct overload to std::string_view. That's great.

https://github.com/jbeder/yaml-cpp/blob/master/include%2Fyaml-cpp%2Femitter.h#L210

liuzicheng1987:
@ZXShady , unfortunately, the string_view support does not appear to be included in the latest version of yaml-cpp. However, I was able to get string_view support in pugixml after upgrading to pugixml 1.15.

PR #448, opened 2025-06-10
Dont repeat class names template arguements

ZXShady:
Repeating ourseleves provides no value and it is verbose


liuzicheng1987:
Merged, thanks

PR #453, opened 2025-07-02
added REFLECTCPP_INSTALL to allow user to control if required

BestITUserEUW:
Allow the user to control installation of reflectcpp libraries. I need it for this lib [kvdb-cpp](https://github.com/BestITUserEUW/kvdb-cppl) if a user wants to build with deps i want to install reflect-cpp too even if not being the top level project


liuzicheng1987:
@BestITUserEUW , thank you!

PR #456, opened 2025-07-08
rename _res to _result to avoid macro conflict

lixin-wei:
I met a macro confilict.

It's in `resolv.h` in linux. In which there is a macro called `_res`.

So I renamed  _res to _result to avoid macro conflict.

BTW, just curious, why do you name all variables with a prefix underline? These names are likely to be used in system headers.

### DIFF
--- a/include/rfl/OneOf.hpp
+++ b/include/rfl/OneOf.hpp
@@ -43,7 +43,7 @@ struct OneOf {
   static rfl::Result<T> validate_impl(const T& _value,
                                       std::vector<Error> _errors) {
     return Head::validate(_value)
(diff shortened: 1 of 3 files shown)


liuzicheng1987:
@lixin-wei thank you for your contribution !

lixin-wei:
@liuzicheng1987 Just curious, may I ask why you name all variables with a prefix underscore? These names are likely to be used in system headers.

ZXShady:
I am personally against using _blah variable names as they are ugly and they conflict with macros just drop the _ and they aren't safe to use in the global namespace.

PR #462, opened 2025-07-13
cbor supports std::optional; fixes #135

### LINKED ISSUE #135: `std::optional` for cbor
The cbor implementation doesn't seem to be handling `std::optional` as documented (for json at least). Consider the following struct example:

```
struct optional_test_t {
    int i;
    std::optional<int> opt_i;
};
```

#### Reading
```
const optional_test_t test =
    optional_test_t{
        .i = 42
    };

std::vector<uint8_t> cbor_bytes = {
    0xA1,      // map(1)
    0x61,      // text(1)
        0x69,  // "i"
    0x18, 0x2A // unsigned(42)
};
const std::string json_string = R"({"i":42})";

auto test_from_json_res = rfl::json::read<optional_test_t>(json_string);
auto test_from_cbor_res = rfl::cbor::read<optional_test_t>((char*)cbor_bytes.data(), cbor_bytes.size());

optional_test_t test_from_json = test_from_json_res.value(); // OK
optional_test_t test_from_cbor = test_from_cbor_res.value(); // Field named 'opt_i' not found.
```

#### Writing

```
const optional_test_t test =
    optional_test_t{
        .i = 42
    };

const std::vector<char> expected_cbor = {
    (char)0xA1,            // map(1)
    (char)0x61,            // text(1)
        (char)0x69,        // "i"
    (char)0x18, (char)0x2A // unsigned(42)
};
const std::string expected_json = R"({"i":42})";

const std::string json = rfl::json::write(test);
const std::vector<char> cbor = rfl::cbor::write(test);

assert(expected_json == json); // OK
assert(expected_cbor == cbor); // Fail. The written cbor bytes translates to "{"i": 42, "opt_i": null}"
```

Just like for `rfl::json`, I expected `rfl::cbor::read()` to successfully parse a buffer of bytes that doesn't contain `std::optional` fields leaving the missing optionals as `std::nullopt`. I also expected `rfl::cbor::write()` to not emit `std::optional` fields in the buffer of bytes.

Long story short, would it be possible for `rfl::cbor` to behave like `rfl::json` when it comes to `std::optional`?

Thanks!

dcorbeil:
I went for the option to generate indeterminate length maps. This allows to not require knowing the size of a struct (for example) in advance for the top level size of the cbor maps or even nested maps.

Eventually adding an option to control whether the generated cbor is determinate or indeterminate map lengths.


tests/cbor/test_integers.cpp
```diff
@@ -18,19 +18,20 @@ TEST(cbor, test_integers_signedness) {
 
     std::vector<char> unsigned_buffer = rfl::cbor::write(Unsigned{BIG_INT});
     std::vector<unsigned char> unsigned_expected = {
-        0xA1, 0x63, 0x75, 0x36, 0x34,
-        0x1B, // Per RFC 8949, Initial byte '0x1B' indicates "unsigned integer (eight-byte uint64_t follows)" 
+        0xBF, 0x63, 0x75, 0x36, 0x34,
+        0x1B, // Per RFC 8949, Initial byte '0x1B' indicates "unsigned integer (eight-byte uint64_t follows)"
               // See: https://www.rfc-editor.org/rfc/rfc8949.html#section-appendix.b
-        0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff
+        0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF
```
bryceschober: Suggestion: Line-break the indefinite length stop code, so it looks more separate from the data, as in your other example:
```suggestion
        0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 
        0xFF
```

tests/cbor/test_optional_fields.cpp
```diff
@@ -0,0 +1,62 @@
...
+struct Person {
+  rfl::Rename<"firstName", std::string> first_name;
+  rfl::Rename<"lastName", std::string> last_name = "Simpson";
+  rfl::Rename<"children", std::optional<std::vector<Person>>> children;
+
+  bool operator==(const Person& other) const {
+    return first_name.get() == other.first_name.get() &&
+           last_name.get() == other.last_name.get() &&
+           children.get() == other.children.get();
+  }
```
bryceschober, line 20: Suggestion: Use the [C++20 default comparison operator](https://en.cppreference.com/w/cpp/language/default_comparisons.html):
```suggestion
  bool operator==(const Person& other) const = default;
```

Although now that I'm thinking about it, perhaps this was done because the default comparison operator is not able to use the `const Type& operator()()` that is [defined](https://github.com/getml/reflect-cpp/blob/18feb020997f61048d3141f98041cf8c964f20be/include/rfl/Rename.hpp#L72) for the `rfl::Rename<Type>` member types... That would be unfortunate.

If that's the case, it would be helpful to briefly document that fact here.

bryceschober (commented):
LGTM, with the exception of some small suggestions.

tests/cbor/test_optional_fields.cpp
dcorbeil: > Although now that I'm thinking about it, perhaps this was done because the default comparison operator is not able to use the `const Type& operator()()` that is [defined](https://github.com/getml/reflect-cpp/blob/18feb020997f61048d3141f98041cf8c964f20be/include/rfl/Rename.hpp#L72) for the `rfl::Rename<Type>` member types... That would be unfortunate.

Yep, exactly

bryceschober: @liuzicheng1987 @dcorbeil I wonder if there is some comparison operator that `rfl::Rename<Type>` could implement to "pass through" default comparison operator support to its type.

liuzicheng1987:
@dcorbeil thanks for the PR. Sorry it took me a couple of days to get to it, I was a bit busy. I will take a closer look tonight.

liuzicheng1987:
@dcorbeil thank you for your contribution! @bryceschober , thanks for reviewing it!

dcorbeil:
@liuzicheng1987 My pleasure. Thanks for the great library!

PR #464, opened 2025-07-16
Fix bug reading with 'tagged' variants when a variant is non-copyable

jmcken8:
Added a test. Needed to change the test to emplace each element rather than initializer_list due to the non-copyable element.


liuzicheng1987:
@jmcken8 , makes sense. Thank you for the contribution!

PR #467, opened 2025-07-18
Allow using `rfl::AddTagsToVariants` with `rfl::Generic` and same-name-different-namespace structs

jmcken8:
The first one was the main reason for this change; the second one (ex: `a::Message` and `b::Message`) was a positive result of applying the change. Added tests for both.

The issue was that both `std::vector<Generic>` and `rfl::Object<Generic>` were getting the type name 'Generic', which caused a compiler error due to duplicated field names.

This will slightly increase the size of serialized formats when using the AddTagsToVariants, but it seems like a worthwhile change.


bryceschober:
@jmcken8 IMO, the additional size needed for namespaces, which are often nested, means that this change should be of an opt-in variety, as in `rfl::AddNamespacedTagsToVariants`. There are serialization formats here that are designed for much more space-constrained contexts than JSON.

jmcken8:
@bryceschober I'm on the fence - AddTagsToVariants already seems like an edge-case issue where it really should *just work* no matter what's given to it, especially since there are several available ways to add tags explicitly (rfl::TaggedUnion, rfl::Variant with rfl::Field) if a user is worried about the serialized size.

What would you suggest for the `rfl::Generic` case, i.e. it contains a std::variant with `std::vector<rfl::Generic>` -> `Generic` and `rfl::Generic<rfl::Generic>` -> `Generic`? Whether or not namespaces are added, I think a change should be made so `AddTagsToVariants` works with `rfl::Generic`.

bryceschober:
@jmcken8 Well I'm not sure how you would make your solution work either. I'm guessing that you've looked at the test failures reported here by now. Looking at [GCC 12's test failure](https://github.com/getml/reflect-cpp/actions/runs/16376537194/job/46277783173?pr=467#step:10:107), I see:
```
[ RUN      ] json.test_add_tag_to_variant_with_generic
/home/runner/work/reflect-cpp/reflect-cpp/tests/json/write_and_read.hpp:35: Failure
Expected equality of these values:
  json_string1
    Which is: "{\"test_add_tag_to_variant::APIResult\":{\"result\":{\"std::__cxx11::basic_string<char>\":\"200\"}}}"
  _expected
    Which is: "{\"test_add_tag_to_variant::APIResult\":{\"result\":{\"std::string\":\"200\"}}}"
Test failed on write. Expected:
{"test_add_tag_to_variant::APIResult":{"result":{"std::string":"200"}}}
Got: 
{"test_add_tag_to_variant::APIResult":{"result":{"std::__cxx11::basic_string<char>":"200"}}}

```

But those are "internal" namespaces, so they're different on for example [LLVM 17](https://github.com/getml/reflect-cpp/actions/runs/16376537194/job/46277783123?pr=467#step:10:108):
```
/home/runner/work/reflect-cpp/reflect-cpp/tests/json/write_and_read.hpp:35: Failure
Expected equality of these values:
  json_string1
    Which is: "{\"test_add_tag_to_variant::APIResult\":{\"result\":{\"std::basic_string<char>\":\"200\"}}}"
  _expected
    Which is: "{\"test_add_tag_to_variant::APIResult\":{\"result\":{\"std::string\":\"200\"}}}"
Test failed on write. Expected:
{"test_add_tag_to_variant::APIResult":{"result":{"std::string":"200"}}}
Got: 
{"test_add_tag_to_variant::APIResult":{"result":{"std::basic_string<char>":"200"}}}
```

My initial guess is that is the original reason for removing the namespace, and I don't know how you would get to where you want from here.

I don't know your use case's requirements, but you do have the opportunity to differentiate your variant members by using the manual tagging suggested in the docs for variants:
```c++
namespace Result {
struct Message {
  using Tag = rfl::Literal<"Result::Message">;
  std::string result;
};
}  // namespace Result

namespace Error {
struct Message {
  using Tag = rfl::Literal<"Error::Message">;
  std::string error;
  int error_id;
};
};  // namespace Error
```
... or whatever naming method of manual disambiguation you'd like.

I also wonder what would happen if you used more fully qualified aliases for the struct types:
```c++
using ResultMessage = Result::Message;
using ErrorMessage = Result::Message;
using Messages = std::variant<ResultMessage, ErrorMessage>;
```
If that method works, then at least you wouldn't be manually keeping string name around that could get out of sync over time.

liuzicheng1987:
Hi @jmcken8 and @bryceschober , how should we proceed with this PR? Should I take a look?

bryceschober:
Yes, @liuzicheng1987 I think that would be helpful.

jmcken8:
Thanks, @liuzicheng1987, yes, that'd be helpful. @bryceschober that would work for the custom case, but it's really the `variant` case that's the issue, fixing the other one was just a (perhaps unwanted) side-effect.

I haven't had a chance to get back to this in the last month and won't for a few more weeks, so any information you can provide would be helpful @liuzicheng1987.

liuzicheng1987:
@jmcken8 @bryceschober it may not be the most elegant solution I ever came up with, but it gets the job done.

tests/json/test_add_tag_to_rfl_variant.cpp
```diff
@@ -34,6 +32,6 @@ TEST(json, test_add_tag_to_rfl_variant) {
 
   write_and_read<rfl::AddTagsToVariants>(
       vec,
-      R"([{"button_pressed_t":{}},{"button_released_t":{"button":4}},{"key_pressed":{"key":99}},{"int":3}])");
+      R"([{"test_add_tag_to_rfl_variant::button_pressed_t":{}},{"test_add_tag_to_rfl_variant::button_released_t":{"button":4}},{"key_pressed":{"key":99}},{"int":3}])");
 }
```
bryceschober, line 36: Suggestion: Add another test to explicitly demonstrate how to avoid any namespace decoration in the serialization.
```suggestion

// Alternatively, in order to avoid getting the namespace to every alternative in our auto-tagged
// variant, we must manually tag all variant alternatives to control all serialized names

struct button_pressed_tagged_t {
  using Tag = rfl::Literal<"button_pressed">;
};

struct button_released_tagged_t {
  using Tag = rfl::Literal<"button_released">;
  rfl::Box<int> button;
};

using my_tagged_event_type_t =
    rfl::Variant<button_pressed_tagged_t, button_released_tagged_t, key_pressed_t, int>;

TEST(json, test_add_manual_tags_to_rfl_variants) {
  std::vector<my_tagged_event_type_t> vec;
  vec.emplace_back(button_pressed_tagged_t{});
  vec.emplace_back(button_released_tagged_t{rfl::make_box<int>(4)});
  vec.emplace_back(key_pressed_t{'c'});
  vec.emplace_back(3);

  write_and_read<rfl::AddTagsToVariants>(
      vec,
      R"([{"button_pressed":{}},{"button_released":{"button":4}},{"key_pressed":{"key":99}},{"int":3}])");
}

```

tests/json/test_add_tag_to_variant.cpp
```diff
@@ -22,12 +23,54 @@ struct key_pressed_t {
...
+// once 'remove_namespaces' is applied to the type name extraction (which is removed)
+// in the MR this test is added
+struct APIResult {
+  rfl::Generic result;
+};
+struct APIError {
+  rfl::Generic error;
+};
+
+using APICallOutput = rfl::Variant<APIResult, APIError>;
```
bryceschober, line 37: Please fix: I think this was meant to be `std::variant`:

```suggestion
using APICallOutput = std::variant<APIResult, APIError>;
```

tests/json/test_add_tag_to_variant.cpp
```diff
@@ -22,12 +23,54 @@ struct key_pressed_t {
...
+
+// test 3 -> two structs with the same name in different namespaces should still
+// be serializable
+
+namespace Result {
+struct Message {
+  std::string result;
```
bryceschober, line 44: Suggestion: Make a note of the alternative approach using `rfl::Tag`:
```suggestion
  // Alternatively, manually tag instead of getting the namespaced name:
  // rfl::Tag<"result_message">
  std::string result;
```

tests/json/test_add_tag_to_variant.cpp
```diff
@@ -22,12 +23,54 @@ struct key_pressed_t {
...
+};
+}  // namespace Result
+
+namespace Error {
+struct Message {
+  std::string error;
```
bryceschober, line 50: Suggestion: Make a note of the alternative approach using `rfl::Tag`:
```suggestion
  // Alternatively, manually tag instead of getting the namespaced name:
  // rfl::Tag<"error_message">
  std::string error;
```

tests/json/test_add_tag_to_variant.cpp
bryceschober, line 1: Suggestion: Overall, I think it would be helpful to maintain the test structure of `test_add_tag_to_variant.cpp` and `test_add_tag_to_rfl_variant.cpp` in parallel as much as possible, because otherwise they kinda imply that they have different functionality.

liuzicheng1987:
@bryceschober I like your second proposal of simply having two separate processors. It would also be a non-braking change, which is always good.

bryceschober:
> @bryceschober I like your second proposal of simply having two separate processors. It would also be a non-braking change, which is always good.

@liuzicheng1987, I have published #494 as my alternative for your and @jmcken8 to review.

liuzicheng1987:
@jmcken8 @bryceschober I am going to close this, because the issue was solved in another PR.

PR #469, opened 2025-07-22
Add support for reading from a byte-container for cbor and ubjson

Ian-Flury:
- Adds a `ContiguousByteContainer` concept.
- Adds support in the cbor and ubjson read functions for reading from a `ContiguousByteContainer`.
- I'm not sure whether it's better to leave the concepts more generic, or build them as wrappers around the `jsoncons` library's type traits, maybe in `include/thirdparty/jsoncons_config.hpp`.


liuzicheng1987:
@Ian-Flury thanks for your contribution.

PR #476, opened 2025-08-05
Make sure that rfl::Underlying enum behaves as expected; fixes #463

### LINKED ISSUE #463: Json Read rfl::UnderlyingEnum processor is broken
Branch: main

Problem: rfl::UnderlyingEnum Processor in rfl::json::read parses anything you give it.
Compiler: gcc-14 / clang-20

Reproducible example:
```cpp
#include "rfl/json/read.hpp"
#include "rfl/UnderlyingEnums.hpp"

enum class CameraSource : uint8_t { kWide, kInfrared, kZoom };
struct ChangeCameraSourcePayload {
    CameraSource source;
};

constexpr const char* bad_payloads[2] = {R"({"source":20})", R"({"source":"hello"})"};

template <class... Processors>
void RunTest() {
    for (auto bad_payload : bad_payloads) {
        auto result = rfl::json::read<ChangeCameraSourcePayload, Processors...>(bad_payload);
        if (!result) {
            printf("Works as expected\n");
            continue;
        }
        printf("Unexpected result: %d\n", static_cast<int>(result->source));
    }
}

auto main() -> int {
    RunTest<rfl::UnderlyingEnums>();
    RunTest();
    return 0;
}
```

Build:
```bash
g++ test_enum.cpp -std=gnu++20 -lreflectcpp -L<lib_dir> -I<include_dir>
```

Result:
```bash
./a.out 
Unexpected result: 20
Unexpected result: 112
Works as expected
Works as expected
```

Same happens when you let the compiler decide the underlying type of the enum. Weirdly both values are consistent if being run multiple times.

liuzicheng1987:


ZXShady:
is there a specific reason for restricting it to only scoped enums? serms arbitrary to me

I also made enchantum available on vcpkg, however it is version 0.3.0 and i am currenty developing version 0.4.0 which comes with alot of things most importantly generators (binary size optimizations basically ) and bound checking.

PR #479, opened 2025-08-12
Add interfaces for jwt-cpp compatibility

aligirayhanozbay:
I want to extend the [jwt-cpp library](https://github.com/Thalhammer/jwt-cpp) so it has reflect-cpp compatibility. The interfaces exposed by reflect-cpp are already almost compatible with the traits expected by jwt-cpp, but a small addition to rfl::Object is needed so that it exposes a `std::size_t count(const key_type& key) const` method. This PR adds that methd.


liuzicheng1987:
@aligirayhanozbay thanks for the contribution!

PR #484, opened 2025-08-19
Better forwarding to improve read performance a little

GregTheMadMonk:
Hello!

This isn't really a significant change, but it helps speed up `rfl::json::read<rfl::Generic>` almost twice in the release build (at least in the benchmarked case).

Sorry for not using your benchmark suite here, I had problems getting the `vcpkg` dependencies properly compiled with GCC15 that ships with my system.

I also tried to look into improving the debug build performance of `rfl::to_generic`, but couldn't find anything to improve there, at least not as simple.

### In release mode

Before:
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
read_struct               95.5 ns         94.4 ns      7332806
read_generic               645 ns          637 ns      1117257
write_struct               104 ns          103 ns      6884743
write_generic              144 ns          142 ns      4990100
to_generic                 115 ns          114 ns      6160175
to_generic_via_json        774 ns          765 ns       932490
```

After:
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
read_struct               94.9 ns         93.7 ns      7444236
read_generic               367 ns          363 ns      1945679
write_struct               104 ns          103 ns      6845176
write_generic              132 ns          130 ns      5413376
to_generic                 119 ns          118 ns      5871778
to_generic_via_json        497 ns          491 ns      1442071
```

### In debug mode

Before:
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
read_struct               1461 ns         1445 ns       493132
read_generic              9450 ns         9337 ns        74690
write_struct               657 ns          649 ns      1092616
write_generic             1083 ns         1072 ns       646018
to_generic                3980 ns         3937 ns       175912
to_generic_via_json      10213 ns        10119 ns        69557
```

After:
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
read_struct               1485 ns         1465 ns       483753
read_generic              7859 ns         7753 ns        91012
write_struct               665 ns          659 ns      1067324
write_generic             1073 ns         1063 ns       660308
to_generic                3925 ns         3899 ns       177444
to_generic_via_json       8641 ns         8539 ns        81955
```

### Benchmark code

```c++
#include <benchmark/benchmark.h>

#include <string>

#include <rfl.hpp>
#include <rfl/json.hpp>

constexpr auto jsd = R"json({ "str": "hello", "i": 42 })json";

struct S { std::string str; int i; };

const S s_etalon{ .str = "hello", .i = 42 };
const rfl::Generic::Object g_etalon = [] () {
    rfl::Generic::Object ret;
    ret["str"] = "hello";
    ret["i"]   = 42;
    return ret;
} ();

using rfl::json::read;
using rfl::json::write;

void read_struct(benchmark::State& state) {
    for (auto _ : state) {
        benchmark::DoNotOptimize( read<S>(jsd).value() );
    }
} // <-- void read_struct(state)

void read_generic(benchmark::State& state) {
    for (auto _ : state) {
        benchmark::DoNotOptimize( read<rfl::Generic::Object>(jsd).value() );
    }
} // <-- void read_generic(state)

void write_struct(benchmark::State& state) {
    for (auto _ : state) {
        benchmark::DoNotOptimize( write(s_etalon) );
    }
} // <-- void write_struct(state)

void write_generic(benchmark::State& state) {
    for (auto _ : state) {
        benchmark::DoNotOptimize( write(g_etalon) );
    }
} // <-- void write_generic(state)

void to_generic(benchmark::State& state) {
    for (auto _ : state) {
        benchmark::DoNotOptimize( rfl::to_generic(s_etalon) );
    }
} // <-- void to_generic(state)

void to_generic_via_json(benchmark::State& state) {
    for (auto _ : state) {
        benchmark::DoNotOptimize( rfl::json::read<rfl::Generic::Object>(rfl::json::write(s_etalon)).value() );
    }
} // <-- void to_generic(state)

BENCHMARK(read_struct);
BENCHMARK(read_generic);
BENCHMARK(write_struct);
BENCHMARK(write_generic);
BENCHMARK(to_generic);
BENCHMARK(to_generic_via_json);

BENCHMARK_MAIN();
```

### DIFF
--- a/include/rfl/Result.hpp
+++ b/include/rfl/Result.hpp
@@ -21,10 +21,13 @@ namespace rfl {
 class Error {
  public:
   Error(const std::string& _what) : what_(_what) {}
+  Error(std::string&& _what) : what_(std::move(_what)) {}
 
   Error(const Error& e) = default;
+  Error(Error&& e) = default;
 
   Error& operator=(const Error&) = default;
+  Error& operator=(Error&&) = default;
 
   /// Returns the error message, equivalent to .what() in std::exception.
   const std::string& what() const { return what_; }
@@ -429,6 +432,10 @@ class Result {
 /// Shorthand for unexpected error.
 inline Unexpected<Error> error(const std::string& _what) {
   return Unexpected<Error>(Error(_what));
+}
+
+inline Unexpected<Error> error(std::string&& _what) {
+  return Unexpected<Error>(Error(std::move(_what)));
 }
 
 /// Shorthand for unexpected error.
--- a/include/rfl/parsing/MapReader.hpp
+++ b/include/rfl/parsing/MapReader.hpp
@@ -55,11 +55,13 @@ class MapReader {
     }
   }
 
-  Result<std::pair<KeyType, ValueType>> make_key(auto& _pair) const noexcept {
+  Result<std::pair<KeyType, ValueType>> make_key(auto&& _pair) const noexcept {
+    static_assert(std::is_rvalue_reference_v<decltype(_pair)>, "Expected an rvalue");
     const auto to_pair =
-        [&](auto _key) -> Result<std::pair<KeyType, ValueType>> {
+        [&](auto&& _key) -> Result<std::pair<KeyType, ValueType>> {
       try {
-        return std::make_pair(KeyType(std::move(_key)),
+        using K = decltype(_key);
+        return std::make_pair(KeyType(std::forward<K>(_key)),
                               std::move(_pair.second));
       } catch (std::exception& e) {
         return error(e.what());
@@ -77,7 +79,7 @@ class MapReader {
                     std::is_floating_point_v<ReflT>) {
         return key_to_numeric<ReflT>(_pair).and_then(to_pair);
       } else {
-        return to_pair(_pair.first);
+        return to_pair(std::move(_pair.first));
       }
 
     } else {
@@ -89,7 +91,7 @@ class MapReader {
       const std::string_view& _name, const InputVarType& _var) const noexcept {
     const auto to_pair = [&](ValueType&& _val) {
       auto pair = std::make_pair(std::string(_name), std::move(_val));
-      return make_key(pair);
+      return make_key(std::move(pair));
     };
     return Parser<R, W, std::remove_cvref_t<ValueType>, ProcessorsType>::read(
                *r_, _var)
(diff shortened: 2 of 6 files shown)


liuzicheng1987:
@GregTheMadMonk that sounds great, but some of the tests appear to be failing. Could you take a look?

GregTheMadMonk:
@liuzicheng1987 if you're talking about the compilation errors (`-Werror=maybe-uninitialized`), then I'm honestly kind of out of ideas for what may cause it, and I also observed it with `main` too. I had to add `-w` to my `CMAKE_CXX_FLAGS` in order to compile tests locally.

The "offending" piece of code in `NamedTupleParser.hpp`:
```c++
  static std::optional<Error> read_object_or_array_with_default(
      const R& _r, const InputObjectOrArrayType& _obj_or_arr,
      NamedTupleType* _view) noexcept {
    std::vector<Error> errors;
    const auto reader = ViewReaderWithDefaultType(&_r, _view, &errors);
    std::optional<Error> err; // <-- declaration here
    if constexpr (_no_field_names) {
      err = _r.read_array(reader, _obj_or_arr);
    } else {
      err = _r.read_object(reader, _obj_or_arr);
    }
    if (err) {
      return err; // <-- -Werror=maybe-uninitialized - why?!
    }
    if (errors.size() != 0) {
      return to_single_error_message(errors);
    }
    return std::nullopt;
  }
};
```
(and a similar function above it) appears safe and straightforward to me... I tried to reproduce the error in a similar scenario and didn't succeed: https://godbolt.org/z/4scKPvnnM (this also didn't produce the same `-Werror` for me locally)

My only two ideas would be either a) GCC gets confused and it's a misfire of this warning or b) there is something in the particular implementations of `.read_object`/`.read_array` used that perform a maybe-uninitialized access to the data which GCC sees but doesn't report as the operation that set the optional into that potentially incorrect state

liuzicheng1987:
Ok. Could you try to just set err = std::nullopt when you first declare it?

GregTheMadMonk:
It didn't help, sadly, for me at least. Same with replacing `if (err)` with `if (err.has_value())`:
image

liuzicheng1987:
@GregTheMadMonk , this should fix the issue. We'll wait for the tests to run through and then this is ready to be merged.

Thank you for the contribution!

GregTheMadMonk:
Thank you!

How did you come to this fix? Or was that the warning system in GCC has actually misfired and you had to do it by trial and error?

liuzicheng1987:
@GregTheMadMonk I had figured that the problem only occurs where we have a mutable optional variable, moreover I recall having seen something like this before. So that's how I figured it out.

PR #494, opened 2025-09-09
Add `rfl::AddNamespacedTagsToVariants` for `rfl::Generic` and same-name-different-namespace structs

bryceschober:
As an alternative to #467, I offer this draft proposal.

Details:

- Add a separate `rfl::AddNamespacedTagsToVariants` processor to support the usage of `rfl::Generic` directly in auto-tagged variants.
- Reuse code paths for `rfl::AddTagsToVariants` by adding a defaulted non-type template parameter to `rfl::VariantAlternativeWrapper`
- Build a set of robust and parallel test suites for all 4 combinations of variant auto-tagging
- Rename the previously-existing variant-tagging test files to better match the processor name.

Outstanding items:

- [x] Update documentation to cover the new functionality and use-cases.
- [ ] Is adding the parallel processor's functionalty to the existing `AddTagsToVariants.hpp` and `is_add_tags_to_variants_v.hpp` acceptable? Those files were named specifically matching the old processor, and now I've added to them.

### DIFF
--- a/tests/cbor/test_integers.cpp
+++ b/tests/cbor/test_integers.cpp
@@ -1,6 +1,7 @@
 #include <gtest/gtest.h>
 
 #include <rfl/cbor.hpp>
+#include "write_and_read.hpp"
 
 namespace test_integers {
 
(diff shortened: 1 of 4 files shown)


include/rfl/Processors.hpp
```diff
@@ -38,6 +39,10 @@ struct Processors<Head, Tail...> {
       std::disjunction_v<internal::is_add_tags_to_variants<Head>,
                          internal::is_add_tags_to_variants<Tail>...>;
 
+  static constexpr bool add_namespaced_tags_to_variants_ =
```
bryceschober, line 42: FYI, at one point I considered setting both `add_tags_to_variants_` and `is_add_namespaced_tags_to_variants` for the new `rfl::AddNamespacedTagsToVariants` to try to simplify some shared code-paths, but decided against that, to keep it distinct from the other processors.

liuzicheng1987: @bryceschober , I agree with your approach. I think we should simply insert a static_assert to make sure that people aren't adding both processors at the same time.

liuzicheng1987: @bryceschober , I tried doing it myself, but apparently I can't push into the PR. No problem, here's the idea:

```cpp
  template <class T, class NamedTupleType>
  static auto process(NamedTupleType&& _named_tuple) {
    static_assert(!add_tags_to_variants_ || !add_namespaced_tags_to_variants_,
                  "You cannot add both rfl::AddTagsToVariants and "
                  "rfl::AddNamespacedTagsToVariants.");
    return Processors<Tail...>::template process<T>(
        Head::template process<T>(std::move(_named_tuple)));
  }
```

liuzicheng1987:
@bryceschober , other than the one comment I made, which you can basically just copy + paste, I like this very much. This is basically ready to be merged.

bryceschober:
@liuzicheng1987 I've implemented your requested `static_assert()`.

I also added CBOR test coverage representative of our use-case, and using a new method of detailed CBOR test expectations that I think will be more practical going forward. It kinda sucks to depend on outside tooling like https://cbor.me, but it's also relatively easy to use and help connect the dots with the test failure formatter I implemented.

liuzicheng1987:
@bryceschober thank you!

jmcken8:
Looks good to me, thanks for doing this work @bryceschober

Issue #429, opened 2025-05-05
Support CBOR parsing using error codes


bryceschober:
It appears to me that all CBOR parsing failures are now requiring exceptions, since #325 was merged. It appears to me that [all of the current `decode_cbor()` implementations](https://github.com/danielaparker/jsoncons/blob/master/include/jsoncons_ext/cbor/decode_cbor.hpp) require exceptions or aborts. The underlying `decode()` and `read()` functions do at least have error-code-oriented variants, but in the `decode_cbor()` implementations, those are all turned into exceptions, AFAICT.

Without digging more deeply than my C++ expertise would easily go, I can't tell which of those are used by `reflect-cpp`. I also can't discern how much effort it might be for `reflect-cpp` to use the underlying error-code-oriented methods.

But as is, the `Result` and monadic features of `reflect-cpp` aren't really available to CBOR users.

liuzicheng1987:
I think that out CBOR wrapper uses monadic error handling just like all the other formats. 

You can check out include/cbor/Reader.hpp and you will find that we do a lot of checks using the exact same logic we use for the other formats. You will also find that there is frequent use of noexcept, meaning any exception would not be propagated. We also sometimes use a try...except logic which would catch CBOR exceptions.

Or is there a specific situation where the current set up is creating difficulties?

bryceschober:
I will see if I can trim down my private use-case, in switching from about v0.15.0 to latest, into a public test case PR.

bryceschober:
See my draft PR #432. I don't know how to work out the correct solution, but I committed a hacky-crap "sledgehammer" approach.

liuzicheng1987:
@bryceschober , in my view your "hacky-crap sledgehammer" approach is totally fine. I see nothing wrong with it.

bryceschober:
> [@bryceschober](https://github.com/bryceschober) , in my view your "hacky-crap sledgehammer" approach is totally fine. I see nothing wrong with it.

My "problem" with it is that jsoncons is [generating a `std::error_code` code](https://github.com/danielaparker/jsoncons/blob/e6d3bc1d4ad70bdf9c8c2831590a5c144e5cb11b/include/jsoncons_ext/cbor/decode_cbor.hpp#L160C9-L160C24), then in the interface that we're using, it turns it into an exception, and then we turn it back into an error object, though not of the "standard" variety.

[...]

PR #497, opened 2025-09-12
Update vcpkg

liuzicheng1987:


bryceschober:
@liuzicheng1987 I'm curious if you're planning on updating the `jsoncons` dependency to 1.4.0, per the needs of #429, or at least have a test build configuration that selects it. I don't know how testing different dependency versions would look on the combination of Github Actions matrix builds & vcpkg...

liuzicheng1987:
@bryceschober OK, will do.

liuzicheng1987:
@bryceschober , I have now updated jsoncons to 1.4.0. It will be part of the 0.21.0 release, which I will draft this weekend.

Issue #504, opened 2025-09-24
[Proposal] Error.what() rvalue move return


BestITUserEUW:
Motivation:

I have a custom error class which currently has to copy Error::what_ which can get expensive with large error messages.

Possible Solution:
```
std::string what() && { return std::move(what_); }
```

Example:
```cpp
class Error {
public:
    Error(const std::string& _what)
        : what_(_what) {}

    Error(const Error& e) = default;

    Error& operator=(const Error&) = default;

    /// Returns the error message, equivalent to .what() in std::exception.
    const std::string& what() const { return what_; }
    
    /// Only qualifies if Error is a rvalue reference
    std::string what() && { return std::move(what_); }

private:
    /// Documents what went wrong
    std::string what_;
};

struct MyLibError {
    std::string message;
    //....
};

auto ParseSomething() -> MyLibError {
    auto result = Error("Parsing failed");  // rfl error message can become really long if many fields are missing
                                            // making a copy expensive

    return MyLibError(std::move(result).what());  // Moved instead of expensive copy
}
```

liuzicheng1987:
Looks good to me. Feel free to open a PR.

BestITUserEUW:
@liuzicheng1987 Sure ill make a pr on the weekend. I want to make some more research on [ref-qualifiers](https://en.cppreference.com/w/cpp/language/member_functions.html#:~:text=Member%20functions%20with%20ref%2Dqualifier) as i have not used them yet, but seems to fit here.

PR #505, opened 2025-09-28
added explicit ref-qualifiers for Error::what() and a additional over…

BestITUserEUW:
…load to allow moving Error::what_ to caller if Error is an rvalue. Adresses #504


BestITUserEUW:
@liuzicheng1987 Do you want me to setup tests for Error::what() ?

liuzicheng1987:
@BestITUserEUW , no, I think that would be overkill. Thanks for your contribution!

Issue #520, opened 2025-10-13
`reflect-cpp` does not generate `.lib` file when building as `shared` on Windows


ErniGH:
When building the library on Windows in shared mode, no symbols are exported and the corresponding .lib file is not generated.

I built it using Conan, and as you can see here the .lib file is not generated:

````
C:/Users/ernesto/.conan2/p/b/refle6511ce2ce100d/p/lib/cmake/reflectcpp/reflectcpp-exports-release.cmake

reflect-cpp/0.19.0: package(): Packaged 1 '.dll' file: reflectcpp.dll
reflect-cpp/0.19.0: package(): Packaged 376 '.hpp' files
reflect-cpp/0.19.0: package(): Packaged 1 '.h' file: yyjson.h
reflect-cpp/0.19.0: package(): Packaged 1 file: LICENSE
reflect-cpp/0.19.0: Created package revision 09f5cceb5cba84250c48638d6b04e5e5
reflect-cpp/0.19.0: Package 'a5d8167e60ca3058a0934e27117193e1e9ea29c2' created
reflect-cpp/0.19.0: Full package reference: reflect-cpp/0.19.0#613991dd09c0fb51a26864aab6ffed46:a5d8167e60ca305
```
Here you can find the full log:
[reflect-cpp.log](https://github.com/user-attachments/files/22882415/reflect-cpp.log)

BestITUserEUW:
@ErniGH Shared libraries on windows are called .dll and .lib are static ones. So that there is no .lib created makes sense as you requested a shared build. If you really want a static library set REFLECTCPP_BUILD_SHARED=OFF when running cmake.

[static libraries](https://en.wikipedia.org/wiki/Static_library)

ErniGH:
Hello, thank you for your response. When you dynamically link a library in Windows, you don't just need the DLLs — Windows also uses the `.lib` file to link the application. In both shared and static linking, .lib files are used: in static, they contain the library itself, and in shared, they assist with the linking process.

```
In load-time dynamic linking, an application makes explicit calls to exported DLL functions like local 
functions. To use load-time dynamic linking, provide a header (.h) file and an import library (.lib) file 
when you compile and link the application. When you do this, the linker will provide the system with 
the information that is required to load the DLL and resolve the exported DLL function locations at
load time.
```
https://learn.microsoft.com/en-us/troubleshoot/windows-client/setup-upgrade-and-drivers/dynamic-link-library#load-time-dynamic-linking

[...]

PR #525, opened 2025-10-16
Feature/shared exports msvc

BestITUserEUW:
This is only a draft for now. This PR addresses #520

@liuzicheng1987 Do you want any changes with the naming or the header file name ? 

Apart from this i wanted to add a shared msvc pipeline but i ran into issues with the dll in the test executables not being on path. 

Do you by any chance know an easy fix to this ? 

Possible solutions i found are putting the build folder on path for all the test exes which i failed to do so. 

Create a cmake helper function that copies everything that matches .dll from build folder into the test exes folder but that didn't feel right. (reflectcpp.dll is not enough as this again tries to load dependent dlls)

### DIFF
--- a/include/rfl/thirdparty/yyjson.h
+++ b/include/rfl/thirdparty/yyjson.h
@@ -30,7 +30,7 @@
 #define YYJSON_H
 
 /*==============================================================================
- * Header Files
+ * MARK: - Header Files
  *============================================================================*/
 
 #include <float.h>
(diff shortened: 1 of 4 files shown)


BestITUserEUW:
@liuzicheng1987 Is there a specific reason that only avro static library is linked when using vcpkg ? 

Linking a static library into a shared one complicates things.

Avro issue could be fixed by linking public to avro and all it's dependants as we are linking manually + using [whole archive](https://learn.microsoft.com/en-us/cpp/build/reference/wholearchive-include-all-library-object-files?view=msvc-170) like the following:

```cmake
target_link_libraries(reflectcpp 
                    PUBLIC 
                        $<LINK_LIBRARY:WHOLE_ARCHIVE,${_AVRO_STATIC_LIB}> 
                        "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/lib/snappy${CMAKE_STATIC_LIBRARY_SUFFIX}" 
                        "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/lib/lzma${CMAKE_STATIC_LIBRARY_SUFFIX}" 
                        "${VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/lib/jansson${CMAKE_STATIC_LIBRARY_SUFFIX}"
)
```

This essentially means that the consumer of this library also has to link against these.

or providing an abstraction layer for avro and use it solely in compilation units. So avro does not need to be linked against by consumers

or just get a shared build of it

BestITUserEUW:
Sources for changes i applied:

Avro dependency list: [avro-c](https://vcpkg.link/ports/avro-c)
Custom copy command from cmake: [Target Runtime Dlls](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#:~:text=TARGET%5FRUNTIME%5FDLLS)
Link whole archive: [Cmake Whole Arhive](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#:~:text=%24%3CLINK%5FLIBRARY%3Afeature%2Clibrary%2Dlist)

BestITUserEUW:
@liuzicheng1987 Ready for review. In the shared tests you can see that .lib file is generated correctly and also verifies that the linkage for all implementations work and they are runnable.

```text
 Generating Code...
     Creating library D:/a/reflect-cpp/reflect-cpp/build/Release/reflectcpp.lib and object D:/a/reflect-cpp/reflect-cpp/build/Release/reflectcpp.exp
  reflectcpp.vcxproj -> D:\a\reflect-cpp\reflect-cpp\build\Release\reflectcpp.dll
```

As there was no answer to my previous question i took the most non-intrusive solution for avro. Make it public link the whole archive and also it's dependants.

Sources for my actions can be found in the comment above

BestITUserEUW:
Once reviewed i will do a merge with main

liuzicheng1987:
@BestITUserEUW , thank you so much for your effort!

PR #537, opened 2025-11-02
Speed up the reflection

liuzicheng1987:


ZXShady:
Any tests that proved it actually improved it?

liuzicheng1987:
@ZXShady , I have done a lot of profiling with Clang's `-f-time-trace` and to be honest, none of this is going to move the needle very much. You can replace all of the structs by rfl::NamedTuple thus completely eliminating the need for any struct reflection and there wouldn't be much of a change in compile time.

The bottleneck isn't the reflection, it is generating the code for serialization and deserialization. We have already done a lot of optimization there, such as implementing a non-recursive tuple (rfl::Tuple) class rather than using the recursively implemented std::tuple. glaze does something similar, but they use a third-party tuple implementation rather than their own. Same idea nonetheless.

However, I liked your idea of nesting the field extractions as a binary tree of constexpr if. It's easy to do and a clear win for larger structs. So I thought why not do it.

liuzicheng1987:
@ZXShady , I will show you what I mean. This is the profile for the test for the README function. The total duration of the struct reflection is about 92 milliseconds, or less than 5% of the total compile time.

image

liuzicheng1987:
@ZXShady , here is the profile for the `test_many_fields` - test, which is a struct with 70 float fields. 

This is obviously a construed example that is unlikely to ever occur in the real world. If you are going to have structs with that many fields, the fields are probably going to be more complex than just floats. In other words, this is probably the greatest share of time spent on struct reflection in any of our tests. 

image

But even for this construed example...we talking about 20-25% of the total compile time. If you can cut that in half, you will have reduced the overall compile time by 10-12% or so. 

But this is really an upper bound...in the real world it's going to be a lot less.

ZXShady:
I wish github mobile actually worked and I got notifications when you pinged me.....

PR #538, opened 2025-11-02
Avoid assuming that "std::string_view" is nullterminated. Generally it is really not.

ActuallyaDeviloper:
Hello, when writing my own parser (reader/writer) support, I noticed that the library incorrectly assumes that "string_view"s are nullterminated. They aren't (sometimes). Also the string length is sometimes recomputed when it's not really necessary.

I was gonna report an issue but then I decided to just fix it myself. Only problem is that I cannot test the other backends (which I do not intend to use) locally right now. Unsurprisingly, the installation scripts fail deep inside some third party dependencies. I do not want to debug that.

### DIFF
--- a/include/rfl/capnproto/Writer.hpp
+++ b/include/rfl/capnproto/Writer.hpp
@@ -262,7 +262,7 @@ class RFL_API Writer {
 
  private:
   kj::StringPtr to_kj_string_ptr(const std::string_view& _str) const {
-    return kj::StringPtr(_str.begin(), _str.end());
+    return kj::StringPtr(_str.data(), _str.size());
   }
 
  private:


liuzicheng1987:
@ActuallyaDeviloper, in practice, the string are always derived from here:

https://github.com/getml/reflect-cpp/blob/main/include/rfl/internal/StringLiteral.hpp

So actually, it is OK to assume that they are null-terminated. But where you can increase safety without generating unnecessary copies, I think it is a good idea.

liuzicheng1987:
@ActuallyaDeviloper , some of the tests are failing to compile. Could you have a look?

liuzicheng1987:
@ActuallyaDeviloper , out of curiosity: For which format did you want to write your own parser? Would you be willing to contribute it to the library?

ActuallyaDeviloper:
Hm, when I implemented my own parser this was not necessarily true anymore, as I was pumping my own "string_view" like data into the writer/reader.

IMHO, if the API says it takes a "string_view" but has some special undocumented preconditions then that is very brittle and bad API design.

At the very least it would need to be documented.

The (nice) alternative I see is to use a different type that does guarantee null termination, however I am not hyped about that because then I need to copy my strings in my parser for every backend.

I think the better alternative I is to accept the data copies for now and petition at the libraries of "avro" and "flexbuf" to add "string_view" APIs.

That said, I am all for removing copies. In fact, there is a lot of copies already anyway. Many backends seem to actually copy the string again into their internal data structure even though e.g. the JSON library would provide a non-copy API. Furthermore when trying to input some string "string_view" as value, it too is copied.

ActuallyaDeviloper:
> @ActuallyaDeviloper , out of curiosity: For which format did you want to write your own parser? Would you be willing to contribute it to the library?

By writing a parser, I meant specializing the "Parser" class. :)
It's for serializing our internal data structures which do not/can not use std-containers, so I don't think I can contribute that.

ActuallyaDeviloper:
I fixed the issue hopefully. This might take a couple rounds because unfortunately I can't compile/test most backends locally.

ActuallyaDeviloper:
A crap, YAML is also failing because the API I am using is unreleased:
https://github.com/jbeder/yaml-cpp/blob/a83cd31548b19d50f3f983b069dceb4f4d50756d/include/yaml-cpp/emitter.h#L210.

The [last release](https://github.com/jbeder/yaml-cpp/issues/1360) 0.8 is from 2023 and does not contain that.

liuzicheng1987:
@ActuallyaDeviloper , I agree, it is brittle design.

For Avro I don't really care that your changes introduce overhead. This format is slow AF anyway.

But flexbuf is one of the two fastest formats supported (the other one being msgpack). Once the code fully compiles, we should take a look at the benchmarks to see how much of an impact this makes. Then, we can make a decision how to proceed here.

liuzicheng1987:
> A crap, YAML is also failing because the API I am using is unreleased: https://github.com/jbeder/yaml-cpp/blob/a83cd31548b19d50f3f983b069dceb4f4d50756d/include/yaml-cpp/emitter.h#L210.
> 
> The [last release](https://github.com/jbeder/yaml-cpp/issues/1360) 0.8 is from 2023 and does not contain that.

Yeah, YAML is fine as well. This is the slowest format we have and no one who uses YAML really cares about speed.

ActuallyaDeviloper:
> > A crap, YAML is also failing because the API I am using is unreleased: https://github.com/jbeder/yaml-cpp/blob/a83cd31548b19d50f3f983b069dceb4f4d50756d/include/yaml-cpp/emitter.h#L210.
> > The [last release](https://github.com/jbeder/yaml-cpp/issues/1360) 0.8 is from 2023 and does not contain that.
> 
> Yeah, YAML is fine as well. This is the slowest format we have and no one who uses YAML really cares about speed.

Can you restart the testing?

ZXShady:
Related to #444

honestly my solution is to adopt something like C++29 zstring_view it is 0 cost string view wrapper

PR #556, opened 2025-12-07
Improve code readability by using concepts where appropriate

liuzicheng1987:


ZXShady:
Why not using std::convertible_to?

PR #558, opened 2025-12-16
Improve bad access error for std::expected<T, rfl::Error>

hos-b:
This is a small cosmetic change that improves the error messages of `std::bad_expected_access` exceptions when the library is used with `REFLECTCPP_USE_STD_EXPECTED`.

I also wrote down a small test that verifies the exception type and message.

### DIFF
--- a/tests/generic/test_bad_access_error.cpp
+++ b/tests/generic/test_bad_access_error.cpp
@@ -1,0 +1,40 @@
+#ifdef REFLECTCPP_USE_STD_EXPECTED
+#include <expected>
+#endif
+
+#include <rfl.hpp>
+#include <gtest/gtest.h>
+
+namespace test_exceptions {
+
+struct Person {
+  rfl::Rename<"firstName", std::string> first_name;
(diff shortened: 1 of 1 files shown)


hos-b:
The [cppreference page](https://en.cppreference.com/w/cpp/utility/expected/bad_expected_access) does not mention anything about the program being ill-formed in case of `std::bad_expected_access` specialization. The way it's worded, I'd even say it's expected that people would customize it:

> Implementations are allowed but not required to override what().
>  bad_expected_access<void> is the base class of all other bad_expected_access specializations.

Regarding the comment about `constexpr`, the move constructor of [std::string](https://en.cppreference.com/w/cpp/string/basic_string/basic_string.html) has been `constexpr` since c++20.

liuzicheng1987:
@hos-b thanks for the contribution! Sorry it took me a couple of days to merge it, I was a bit busy these days.

hos-b:
> @hos-b thanks for the contribution! Sorry it took me a couple of days to merge it, I was a bit busy these days.

No problem. Thank you for the review and happy holidays 🙂

PR #574, opened 2026-01-13
Enable building as a native ROS package

TomCC7:
# Description
This PR adds support for building this project directly within a ROS workspace.

# Motivation
Currently, integrating this library requires external package managers like Conan or vcpkg. To simplify the workflow for ROS developers, I have structured the project to function as a standard ROS package. This allows it to be compiled alongside other ROS nodes without introducing additional dependency management tools.


liuzicheng1987:
@TomCC7 thank you!

PR #581, opened 2026-01-25
Make REFLECTCPP_USE_BUNDLED_DEPENDENCIES option consistent with enchantum

DNKpp:
Hello,
I'm currently reviewing whether I can adapt reflect-cpp for the project I'm working on. Unfortunately, the integration does not work as of now, because I'm already depending on enchantum, thus there is an existing integraiton. As reflect-cpp brings its own copy under the `thirdparty` folder, that results in lots of clashes when including `rfl.hpp`.
There already is the option `REFLECTCPP_USE_BUNDLED_DEPENDENCIES` which does switch whether the *thirdparty* folder is set as include directory, and thus conditionally includes the bundled `ctre` and `yyjson`, or an external version.

Currently, `enchantum` includes are always prefixed with the subfolder `thirdparty`, which prevents this mechanism to work.


liuzicheng1987:
@DNKpp , thanks for your PR!

However, the problem is that enchantum is not on Conan. If I accept this PR, it will for sure break our Conan package.

liuzicheng1987:
@DNKpp , it appears to me that the pragmatic solution is to have a separate flag for whether or not we want to used the bundled enchantum.

DNKpp:
@liuzicheng1987 I think a separate option for each bundled dependency would be the best. I'm certain there are already lots of CTRE users out there, while yyjson is probably not that common in the c++ community. The *all or nothing* option is at least something, but I like your suggestion with dedicated switches.

EDIT: Are all these failed pipelines due to the mentioned conan issue?

liuzicheng1987:
@DNKpp , that means we would have to tweak the existing recipes. So I rather just have a separate flag for enchantum.

The difference between CTRE and YYJSON and enchantum is that the former two can be found on all relevant package managers. So if we do not include a bundled version, there is always an alternative. Enchantum, on the other hand , is not found on all relevant package managers, so we need to treat it differently.

liuzicheng1987:
@DNKpp , yes, all of the pipeline failures are related to the Conan issue.

ZXShady:
> @DNKpp , that means we would have to tweak the existing recipes. So I rather just have a separate flag for enchantum.
> 
> The difference between CTRE and YYJSON and enchantum is that the former two can be found on all relevant package managers. So if we do not include a bundled version, there is always an alternative. Enchantum, on the other hand , is not found on all relevant package managers, so we need to treat it differently.

I should get myself to work on Conan, I made vcpkg but haven't updated it to latest version.

liuzicheng1987:
@DNKpp, I am closing this PR due to inactivity and since some of the tests do not pass. If you want to continue your work here, feel free to reopen this PR.

ZXShady:
@liuzicheng1987 

Enchantum was added to conan and updated for vcpkg by 2 awesome contributors

https://github.com/conan-io/conan-center-index/pull/29999

https://github.com/microsoft/vcpkg/pull/51135

PR #582, opened 2026-01-26
Fix some unused parameters warnings

makekryl:


liuzicheng1987:
@makekryl , thanks for the PR.

In these three instances, instead of writing [maybe_unused], could you simply get rid of the variable name of the unused parameter. 

So instead of 

`([maybe_unused] auto&& _var)`

Just type 

`(auto&&)`

That is what we do in the rest of the codebase and it is also more concise.

makekryl:
I thought `[[maybe_unused]]` is preferred, as it was introduced to some other pieces of code in a similar PR (#547), it also seems quite less misleading than unnamed parameters though, IMO `to_schema` would be a little bit confusing without the argument name

I can also remove those `[[maybe_unused]]` too in this PR if you want to stick with unnamed parameters, what do you think?

liuzicheng1987:
@makekryl , sure, go ahead.

liuzicheng1987:
@makekryl , looks great. I will wait for the tests, then merge. Thanks for your contribution!

makekryl:
Thanks for keeping this great project going!

makekryl:
Seems like one job failed due to vcpkg related infra problem :/

liuzicheng1987:
@makekryl , yes, it happens. I'll just rerun the test. No worries.

PR #596, opened 2026-02-03
Allow usage of stateful visitors inside rfl::Variant

### LINKED ISSUE #595: Inconsistency in rfl::Variant visit overloads prevents stateful visitors
I'm using rfl::Variant and I noticed it is currently not possible to use a stateful visitor even when passing it as non-const reference to visit.
Example:
```cpp

struct Type1 {
    std::string name;
    // other fields
};
struct Type2 {
    std::string name;
    // other fields
};

struct StatefulVisitor {
    std::vector<std::string> collected_data;
    template <typename T>
    void operator()(const T& element) {
        using Type = std::decay_t<T>;
        // impl ...
        // all if constexpr here to handle the different types
        collected_data.push_back(element.name);
    }
};

// then somewhere inside a function
rfl::TaggedUnion<Type1,Type2> my_union = ...;
StatefulVisitor visitor;
my_union.visit(visitor); // this fails here
```

I've looked at the Variant.hpp file and found the issue inside the do_visit_... functions.
The library correctly provides multiple overloads to handle both const F& and F& but the internal lambdas visit_one used to perform the expansion alwasy takes the functor as a const F& regardless of the parent function cv-qualifiers.

For example:
```cpp
  template <class F, class ResultType, IndexType... _is>
  void do_visit_with_result(F& _f, std::optional<ResultType>* _result,
                            std::integer_sequence<IndexType, _is...>) {
    auto visit_one = [this]<IndexType _i>(const F& _f,
                                          std::optional<ResultType>* _result,
                                          Index<_i>) {
      if (!*_result && index_ == _i) {
        _result->emplace(_f(get_alternative<_i>()));
      }
    };
    (visit_one(_f, _result, Index<_is>{}), ...);
  }
```

This effectively negates the purpose of having non const F& overloads.
I could submit a pull request if you agree this is a bug/oversight.

FrancescoMerenda:
This PR allows the usage of stateful visitors inside rfl::Variant.

Fixes #595

### DIFF
--- a/include/rfl/Variant.hpp
+++ b/include/rfl/Variant.hpp
@@ -223,7 +223,7 @@ class Variant {
   template <class F, IndexType... _is>
   void do_visit_no_result(F& _f, bool* _visited,
                           std::integer_sequence<IndexType, _is...>) {
-    auto visit_one = [this]<IndexType _i>(const F& _f, bool* _visited,
+    auto visit_one = [this]<IndexType _i>(F& _f, bool* _visited,
                                           Index<_i>) {
       if (!*_visited && index_ == _i) {
         _f(get_alternative<_i>());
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
@FrancescoMerenda thank you for the contribution!

PR #599, opened 2026-02-06
Remove reference to internals in error message for unused values

### LINKED ISSUE #598: Update error message for unused values
First of all, great library; it allows us to simplify our config parsing significantly.

I have one change request: currently when you set `rfl::NoExtraFields` it will warn on unused fields.
This works correctly, but the error message contains the following string:

```
Remove the rfl::NoExtraFields processor or add rfl::ExtraFields to avoid this error message.
```

I can see how this is useful for developers, but we want to display these error message to the users, so that they know exactly why parsing failed. This part of the message is not exactly useful though, because a user can't change any of these internals.

Could this part of the message be removed? I'd be happy to make the change myself. 

For reference:
https://github.com/getml/reflect-cpp/blob/main/include/rfl/parsing/ViewReader.hpp#L145

BugelNiels:
Prevents exposing library internals in the error message when a value is not used (triggered when `rfl::NoExtraFields` is used). 

fixes #598

### DIFF
--- a/include/rfl/parsing/ViewReader.hpp
+++ b/include/rfl/parsing/ViewReader.hpp
@@ -142,8 +142,7 @@ class ViewReader {
       if (!already_assigned) {
         std::stringstream stream;
         stream << "Value named '" << _current_name_or_index
-               << "' not used. Remove the rfl::NoExtraFields processor or add "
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
@BugelNiels thanks for the contribution!

PR #606, opened 2026-02-09
Improve invalid enum error message

### LINKED ISSUE #605: Improve error message when an invalid enum value is provided.
When an invalid enum value is provided in a config, the `string_to_enum` method will simply say:

```
stoi
```

It would be nice if we could improve this error message to be a bit more informative. For example, along the lines of:

```
Invalid enum value: 'bart'. Must be one of [red, green, blue, yellow].
```

I have already implemented this change; I'll link the PR shortly. Let me know if this looks good.

BugelNiels:
Updates the `enum_to_string` function to have a more elaborate error messages. I had to move around the `get_enumerators()` and related functions as one of them was needed to implement this fix (`get_enumerator_array`). I figured this was cleaner than forward declaring it in the same file.

I have also added a few basic tests to verify the error behaviour. Let me know if additional tests are needed here.

Fixes #605

### DIFF
--- a/include/rfl/enums.hpp
+++ b/include/rfl/enums.hpp
@@ -2,6 +2,7 @@
 #define RFL_ENUMS_HPP_
 
 #include <string>
+#include <sstream>
 
 #include "Result.hpp"
 #include "internal/enums/get_enum_names.hpp"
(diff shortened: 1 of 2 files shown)


BugelNiels:
Replaced the `std::ostringstream` with regular strings so that it doesn't need an additional import. The currently failing tests all seem to be due to transient download failures.

liuzicheng1987:
@BugelNiels Merged. Thanks for the contribution!

PR #607, opened 2026-02-09
Set file name and license in RPM CPack options

### LINKED ISSUE #604: Improve RPM name generated by CPack
Currently, when building an RPM of `reflectcpp` with CPack, the resulting name will look something like this:

```
reflectcpp-0.23.0-Linux.rpm
```

However, this is missing some information from the file name (such as `el9` and `x86_64`).

For Debian packages, The `CMakeLists.txt` already sets the following:

```cmake
set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
```

I would propose to also add the following under the RPM-specific configuration:

```cmake
set(CPACK_RPM_FILE_NAME RPM-DEFAULT)
# This is not strictly necessary, but good for completeness
set(CPACK_RPM_PACKAGE_LICENSE "MIT")
```

I tested it and it would generate a file name like this:

```
reflectcpp-0.23.0-1.el9.x86_64.rpm
```

Let me know if this is okay; if so I can create the PR.

---

Another slightly unrelated question, the repo is called `reflect-cpp`, but the resulting packages generated from CMake are called `reflectcpp`. Is there any particular reason why this does not include the hyphen?

BugelNiels:
Fixes #604

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -514,6 +514,10 @@ set(CPACK_DEBIAN_PACKAGE_DEPENDS "")
 
 # RPM specific configuration
 set(CPACK_RPM_PACKAGE_REQUIRES "")
+set(CPACK_RPM_FILE_NAME RPM-DEFAULT)
+set(CPACK_RPM_PACKAGE_LICENSE "MIT")
+set(CPACK_RPM_PACKAGE_LICENSE "MIT")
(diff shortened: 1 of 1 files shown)


liuzicheng1987:
@BugelNiels thanks for the contribution!

PR #624, opened 2026-02-26
Add rfl::Deprecated, which adds deprecated and deprecationMessage to the generated schema field.

AndrewNolte:


liuzicheng1987:
@AndrewNolte , I just realised that this doesn't have documentation. Could you add documentation in a separate PR?

Both in the README.ms and the docs folder...

PR #632, opened 2026-03-10
Fix "Error()" function.

Ra-LM:
Simple fix for a problem I ran into.


liuzicheng1987:
Hi @Ra-LM , thank you for this contribution. 

However, I think this is already covered by https://github.com/getml/reflect-cpp/pull/618, which we will merge soon.

liuzicheng1987:
@Ra-LM apparently it is still relevant. Thanks for the contribution, I am merging.

PR #634, opened 2026-03-22
Fix capnproto add_value_to_map overwriting all entries at index 0

lixin-wei:
## Summary

Fix a bug in capnproto's add_value_to_map, It doesn't use _parent->ix_++ to track the entry index. Instead, it creates a new OutputArrayType each time with ix_ defaulting to 0, so every entry overwrites index 0 in the capnproto list.

## Test plan

- [x] New test `capnproto.test_map_with_string_values` fails before the fix (only last entry survives round-trip)
- [x] New test passes after the fix
- [x] All 46 existing capnproto tests continue to pass

### DIFF
--- a/include/rfl/capnproto/Writer.hpp
+++ b/include/rfl/capnproto/Writer.hpp
@@ -186,10 +186,11 @@ class RFL_API Writer {
   template <class T>
   OutputVarType add_value_to_map(const std::string_view& _name, const T& _var,
                                  OutputMapType* _parent) const {
-    auto entries =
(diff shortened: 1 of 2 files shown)


lixin-wei:
The failed checks seems not related to my change. It's a network issue.

lixin-wei:
@liuzicheng1987 would you pls have look at my recent 2 PRs when you have time?

liuzicheng1987:
@lixin-wei Yes. Sorry I missed them until now.

liuzicheng1987:
@lixin-wei , looks great. Thanks for the contribution.

PR #638, opened 2026-03-24
Delete MR_DESCRIPTION.md

cmnrd:
While browsing the repository I noticed the markdown file `MR_DESCRIPTION.md` which seems to be left over from merging #618. This PR deletes the file as it probably isn't intended to live in the global repository scope and seems to be support material for #618. Please ignore this PR if the file is there intentionally.


liuzicheng1987:
@cmnrd , thanks!

PR #649, opened 2026-04-04
Minor improvements to the Boost Serialization handling

liuzicheng1987:


ron0studios:
I think the inlining fixes the problems with the Boost CICD before so this is great :)

PR #662, opened 2026-05-09
Add unified rfl::Settings::with<>() API for format-specific settings

Centimo:
## Summary

Replaces ad-hoc `with_<field>()` accessors on `csv::Settings` and
`parquet::Settings` with a single uniform `with<>()` API injected by
the `RFL_SETTINGS_OPS(Derived)` macro. Two overloads are provided:

- `with<&T::field>(value)` — pointer-to-member.
- `with<"field">(value)` — field name as string literal.

Both return a copy with the chosen `const` field replaced.

The old per-field setters (`with_delimiter()`, `with_compression()`,
etc.) are kept as `[[deprecated]]` shims pointing at the new API, so
existing user code keeps compiling with a clear migration warning.

## Motivation

The previous design required a hand-written `with_*()` per field,
growing linearly with each new option. The new macro replaces that
boilerplate with a single line and is consistent across formats.

## Const fields

All data members in `csv::Settings` and `parquet::Settings` are now
declared `const`. Settings are intended to be configured once and
consumed; preventing accidental in-place mutation makes that intent
explicit at the type level. `with<>()` returns a fresh copy, so the
configurable surface is unchanged. The macro's PTM overload carries
a `requires const_member_of<FieldPtr, Derived>` clause, so a call
with a non-const field fails at the call site, not deep inside the
implementation.

## Implementation notes

- `field_index_v<T, FieldPtr>` recovers the field index in a
  `consteval` context. On GCC/Clang it uses the existing fake-object
  approach (`&(fake_object<T>().*FieldPtr)` vs structured-binding
  addresses). MSVC refuses to constant-fold a PTM-deref through an
  extern fake object (C2131), so on MSVC a local default-initialised
  `T{}` is constructed inside `consteval` instead. Both paths share
  a single base helper, `field_extractor<T, n>::get_ptr<I>(obj)`.
- `field_index_by_name_v<T, Name>` does the analogous lookup for the
  string-literal overload, walking field names through the existing
  `__PRETTY_FUNCTION__` parser.
- `get_fake_object.hpp` is unchanged from upstream (`static const`
  wrapper member); the temporary `extern const` variant tried during
  development was rolled back once the local-object MSVC path made
  it unnecessary.

## Tests

`tests/cli/test_settings_macro.cpp` covers single-field replace,
chained replace by PTM, chained replace by name, mixed PTM+name
chains, and string-move semantics. Existing csv/parquet tests are
updated to use the new API. CI: all targets green except known-
unrelated infra failures (vcpkg/conan).


liuzicheng1987:
Hi @Centimo, I think this is certainly a step in the right direction, but I am always hesitant about API-breaking changes like this, because they always tend to frustrate people.

As a compromise, could you reinsert the old functions, add a comment in the code that they are deprecated and will be removed at a later date? Since the markdown documentation only includes the new style, most people will start using that. When we add new functions it will only be in the new style.

What do you think about that?

Centimo:
Hi @liuzicheng1987 

> As a compromise, could you reinsert the old functions?

Do you mean the old with_* functions?

liuzicheng1987:
@Centimo, yes. 

I think that people use them in the productive system and when you have to change that unexpectedly, that makes people angry.

So we should keep them for the time being but add a deprecation notice, because I do agree that your approach is better.

Centimo:
@liuzicheng1987 done

liuzicheng1987:
@Centimo, all right, thank you very much for the contribution. I am merging.

Issue #421, opened 2025-04-29
Crash on rfl::json::write in Windows, if std::filesystem::path (or std::string or std::wstring) contains non-latin symbols


9241304:
1. As you know, native narrow encoding for Windows is not utf-8, so when I`m trying to serialize std::string in native Windows encoding (like cp1251) I'm getting crash. In fact, I can't use the same serialization code for Linux and Windows and I can't use rfl with WinAPI narrow functions easily. I think rfl::json::write should consider native narrow encoding.
2. Ok, I thought I can resolve my problems using std::wstring or std::filesystem::path (I got crash for path returned by SHGetFolderPathA). I tried to use SHGetFolderPathW and  std::wstring or std::filesystem::path, but I also got crash! From other side, if I'm constructing std::filesystem::path from utf-8, then rfl::json::write works, but std::filesystem::path variable contains broken path. If I'm converting this path from utf-8 to std::wstring, then I'm getting crash again.

Version https://vcpkg.link/ports/reflectcpp/v/0.19.0/0
Windows any, Visual Studio any

Minimal code to reproduce:
```
std::ifstream infile("cp1251.txt", std::ios::binary);
std::string data;
std::getline(infile, data);
std::filesystem::path path{ data };
auto json{ rfl::json::write(path) };
```

or

```
std::wstring utf8_to_wstring(const std::string& str) 
{
	std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
	return converter.from_bytes(str);
}

...

std::ifstream infile("utf8.txt", std::ios::binary);
std::string data;
std::getline(infile, data);
auto ws{ utf8_to_wstring(data) };
auto json{ rfl::json::write(ws) };
```

[cp1251.txt](https://github.com/user-attachments/files/19968391/cp1251.txt)
[utf8.txt](https://github.com/user-attachments/files/19968392/utf8.txt)

liuzicheng1987:
Thanks for reporting.

liuzicheng1987:
@9241304 could you show me the exact stack trace?

9241304:
std::filesystem::path
Crash is similar to std::string case. Looks like path::string() was called.
```
 	test.exe!std::_Narrow_char_traits<char,int>::length(const char * const _First) Line 430	C++
 	test.exe!std::string::basic_string<char,std::char_traits<char>,std::allocator<char>>(const char * const _Ptr) Line 745	C++
[...]

PR #668, opened 2026-05-14
fix(json): don't crash on non-ASCII std::wstring / std::filesystem::path

### LINKED ISSUE #422: crash on rfl::json::save() while serializing struct with std::wstring conatining cyrilic and japan chars
I have "settings " structure with std::wstring with cyrillic, japan chars in.

struct settingsInt
{
	std::wstring picture_pathname{L"Вітаю_おめでとう"}; 
        //std::string picture_pathname_utf8; //here string as utf8 only for saving. Crashes on save. Reason, probably wstring path with cyrillic or japan chars.
	int int_cluster_matching_width;
	int int_width_divisor;
...
} settings;

Crashing on writing json file using:
rfl::json::save(NewFileName + "_settings.json"), settings, rfl::json::pretty);

For now I turned to std::string and using UTF-8 encoding to save.
Reflect CPP version from appr,. 26.04.2025

Latest VS and Win10

mrgyatso:
## Summary

A struct field of type `std::wstring` or `std::filesystem::path` containing any non-ASCII
character crashes `rfl::json::read` / `rfl::json::write` with `std::terminate`. This is the
crash behind #421 and #422. Both were reported from Windows, but the crash is not
platform-specific: it reproduces on Linux / glibc / g++ just as readily.

Minimal repro (Linux, default locale):

```cpp
struct S { std::wstring w; };
rfl::json::write(S{ .w = L"é" });     // std::length_error, then std::terminate
rfl::json::read<S>(R"({"w":"é"})");   // std::length_error, then std::terminate (read is noexcept)
```

## Root cause

`Parser_wstring.hpp` transcodes via `std::mbsrtowcs` / `std::wcsrtombs` and passes the
result straight into `resize()`:

```cpp
auto len = std::mbsrtowcs(outStr.data(), &ptr, val.size(), &state);
outStr.resize(len);   // len can be (size_t)-1
```

`std::mbsrtowcs` / `std::wcsrtombs` return `(size_t)-1` when the input is not valid in the
current C locale's `LC_CTYPE` encoding. The default `"C"` locale is ASCII-only, so any
non-ASCII byte makes `len` equal to `(size_t)-1`, which becomes `resize(SIZE_MAX)` and
throws `std::length_error`. `read` is declared `noexcept`, so that becomes an immediate
`std::terminate`. `Parser_filepath.hpp` has the same locale dependency through
`path::string()` and the narrow `path` constructor.

## The fix: two independent commits

Commit 1 serializes `std::filesystem::path` as UTF-8. `Parser_filepath` now uses
`path::u8string()` and construction from `std::u8string` instead of the locale-dependent
`path::string()` and narrow constructor. JSON is UTF-8 (RFC 8259), so this conversion is
well-defined and locale-independent. It uses the standard library only, about one
substantive line per function.

Commit 2 converts `std::wstring` with an explicit UTF-8 codec. It adds a small header-only
UTF-8 to `wchar_t` codec that handles both `wchar_t` widths (UTF-32 on Linux and macOS,
UTF-16 on Windows) and rejects malformed input rather than crashing, then rewires
`Parser_wstring` to use it. `read` now returns an `rfl::Error` on malformed UTF-8, and
`write` produces correct UTF-8. The parser itself gets shorter, since the locale juggling
is gone.

Commit 1 stands on its own, and commit 2 is independent of it.

## Also fixes a crash that is not in either issue

#421 and #422 are about `write` and `save`. The `read` path crashes on the same input,
and because `read` is declared `noexcept`, there is no `rfl::Error` path and no
`try`/`catch` a caller can use: it is an unconditional `std::terminate`. Neither issue
mentions this. Both commits fix `read` and `write`.

## A note on scope

I read through the #421 discussion. The genuinely hard part there, auto-detecting the
encoding of a `std::string`, is ambiguous on Windows, and this PR does not touch it. But
`std::wstring` and `std::filesystem::path` are not encoding-ambiguous (they are
well-defined UTF-16 / UTF-32, and the standard's defined path encoding), so a correct,
locale-independent conversion is well-defined, which is what commit 2 does.

If you would rather not take on the codec in commit 2, I am happy to reduce it to just
guarding the `(size_t)-1` return, so `read` returns an `rfl::Error` and `write` emits an
empty string. That kills the crash without adding conversion logic. I led with the full
fix because it actually makes the types usable, but it is your call.

## Tests

Regression tests added to `tests/json/test_wstring.cpp` and `tests/json/test_filepath.cpp`,
covering round-trips with Latin-1, Cyrillic, CJK, Hiragana, and astral-plane content.

Closes #422. Refs #421: this fully addresses its `std::wstring` and `std::filesystem::path`
cases. The separate `std::string` codepage question raised in that thread is intentionally
left out of scope.

### DIFF
--- a/include/rfl/internal/strings/utf8_conversions.hpp
+++ b/include/rfl/internal/strings/utf8_conversions.hpp
@@ -1,0 +1,170 @@
+#ifndef RFL_INTERNAL_STRINGS_UTF8_CONVERSIONS_HPP_
+#define RFL_INTERNAL_STRINGS_UTF8_CONVERSIONS_HPP_
+
+#include <optional>
+#include <string>
+
+namespace rfl::internal::strings {
+
+// Locale-independent conversion between UTF-8 (std::string) and std::wstring.
+//
+// JSON is defined to be UTF-8 (RFC 8259), so the std::string side is always
+// UTF-8. wchar_t is treated as UTF-32 where it is >= 4 bytes (Linux, macOS) and
+// as UTF-16 where it is 2 bytes (Windows). Unlike std::mbsrtowcs / std::wcsrtombs
+// this does not depend on the process's C locale and cannot return an
+// unchecked error sentinel.
+
+namespace utf8_detail {
+
+/// Decodes UTF-8 bytes into Unicode code points. Returns std::nullopt if the
+/// input is not well-formed UTF-8 (truncated, overlong, bad continuation byte,
+/// surrogate code point, or out of range).
+inline std::optional<std::u32string> decode_utf8(const std::string& _str) {
+  std::u32string out;
+  std::size_t i = 0;
+  const std::size_t n = _str.size();
+  while (i < n) {
+    const auto c = static_cast<unsigned char>(_str[i]);
+    char32_t cp = 0;
+    int extra = 0;
+    if (c < 0x80) {
+      cp = c;
+      extra = 0;
+    } else if ((c >> 5) == 0x6) {
+      cp = c & 0x1F;
+      extra = 1;
+    } else if ((c >> 4) == 0xE) {
+      cp = c & 0x0F;
+      extra = 2;
+    } else if ((c >> 3) == 0x1E) {
+      cp = c & 0x07;
+      extra = 3;
+    } else {
+      return std::nullopt;  // invalid leading byte
+    }
+    if (i + static_cast<std::size_t>(extra) >= n) {
+      return std::nullopt;  // truncated multi-byte sequence
+    }
+    for (int k = 1; k <= extra; ++k) {
+      const auto cc = static_cast<unsigned char>(_str[i + k]);
+      if ((cc >> 6) != 0x2) {
+        return std::nullopt;  // bad continuation byte
+      }
+      cp = (cp << 6) | (cc & 0x3F);
+    }
+    static constexpr char32_t mins[] = {0, 0x80, 0x800, 0x10000};
+    if (cp < mins[extra]) {
+      return std::nullopt;  // overlong encoding
+    }
+    if (cp > 0x10FFFF || (cp >= 0xD800 && cp <= 0xDFFF)) {
+      return std::nullopt;  // out of range or surrogate code point
+    }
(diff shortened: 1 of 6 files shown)


liuzicheng1987:
@mrgyatso , thanks for the contribution. This looks great. I was about to do the 0.25.0 release, but I want to squeeze your commit in, so I will delay it for another day.

liuzicheng1987:
@mrgyatso it appears your solution does not work for Windows/MSVC. Could you take a look?

mrgyatso:
Ah yeah I overlooked that. I'll check it out tonight. Give me a bit.

mrgyatso:
@liuzicheng1987 Sorry I took so long, just got home a bit ago. Pushed the fix in 8657c32. Library code is unchanged; this was a test-build issue. MSVC was reading the test sources using the system code page, so non-ASCII in the new `L"..."` / `u8"..."` literals was mis-decoded before the codec ran (expected `\xC3\xA9`, got `\xC3\x83\xC2\xA9`). The fix is `/utf-8` in `tests/CMakeLists.txt`. Verified locally on VS 2022 / MSVC 19.50: without it the two non_ascii tests reproduce the CI failure, with it the full suite passes 340/340.

Can narrow `/utf-8` to `tests/json/CMakeLists.txt` if you'd prefer.

mrgyatso:
Looks like this one failed due to a 502 error upstream with vcpkg, unrelated to the fix. Should work fine on re-run.

liuzicheng1987:
@mrgyatso , yes, unfortunately the build pipeline is a bit unstable. We have too many things running at the same time, so the vcpkg servers tend to block us. It's a known issue and clearly has nothing to do with your PR.

liuzicheng1987:
@mrgyatso , I have merged this. Thankl you for your contribution!

mrgyatso:
> @mrgyatso , I have merged this. Thankl you for your contribution!

@liuzicheng1987 
Happy to help! I sent you a request on LinkedIn by the way. Let's stay connected. Hope to build with you in the future. :)

PR #672, opened 2026-05-29
Generic::to_*() enhancements

kone-tlammi:
- The new version avoids extra copies in case the current Generic
contents do not match the casted-to type.
- Using std::get_if avoids switch-case generation for all the variant types
  when we are only interested in one or two types.


liuzicheng1987:
@kone-tlammi , thanks for the contribution! I am merging.

Issue #396, opened 2025-03-25
[Feature Request] Add support to JSON Schema "default" keyword


caio-porto:
It would be great if we could generate a JSON Schema with [Default](https://json-schema.org/understanding-json-schema/reference/annotations) fields based on the struct initial value.

For instance:

C++ Struct
```
struct Foo {
    bool value1 = false;
    double value2 = 1.0;
}
```

JSON Schema
```
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$ref": "#/definitions/temperature",
    "definitions": {
        "Foo": {
            "type": "object",
            "properties": {
                "value1": {
                    "type": "boolean",
                    "default": false
                },
                "value2": {
                    "type": "number",
                    "default": 1.0
                }
            },
            "required": [
                "value1",
                "value2"
            ]
        }
    }
}
```

postlund:
I'm using the generated schema in sphinx to generate configuration documentation, but I would love to add default values and indicate what fields that are optional. Would this `rfl::DefaultVal` be possible to use to support this?

PR #680, opened 2026-06-08
WIP: respects DefaultVal in schema and adds default value to schema. fixes #679, #396

### LINKED ISSUE #679: rfl::DefaultVal's are still required in the json schema
For the following struct:
```c++
struct Settings {
    rfl::DefaultVal<std::uint16_t> port = 80;
};
```
the following schema is generated
```json
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$ref": "#/$defs/Settings",
    "$defs": {
        "Settings": {
            "type": "object",
            "properties": {
                "port": {
                    "type": "integer"
                }
            },
            "required": [      <-- This is wrong
                "port"
            ]
        }
    }
}
```
Also, if Settings could be default constructed, the json schema generator could do that to add `"default": 80` for the `port` entry to the schema.

autoantwort:
fixes #679, #396

### DIFF
--- a/include/rfl/parsing/NamedTupleParser.hpp
+++ b/include/rfl/parsing/NamedTupleParser.hpp
@@ -188,10 +188,10 @@ struct NamedTupleParser {
    * @param _definitions The map of definitions to add to.
    * @return The schema type.
    */
-  template<typename View>
+  template <typename View = void>
   static schema::Type to_schema(
-      std::map<std::string, schema::Type>* _definitions, View* _view = (void*) nullptr) noexcept {
-
+      std::map<std::string, schema::Type>* _definitions,
+      View* _view = nullptr) noexcept {
     SchemaType schema;
     build_schema(_definitions, &schema, _view,
                  std::make_integer_sequence<int, size_>());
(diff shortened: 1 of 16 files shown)


include/rfl/parsing/NamedTupleParser.hpp
```diff
@@ -226,14 +229,30 @@ struct NamedTupleParser {
...
       auto s = Parser<R, W, U, ProcessorsType>::to_schema(_definitions);
+      static_assert(std::is_same_v<decltype(s), schema::Type>, "was sonst?");
+      if constexpr (!std::is_same_v<View, void>) {
+
+        if constexpr (std::is_same_v<decltype(s), schema::Type>) {
+          // s.default_value_ = Parser<R, W, U, ProcessorsType>::write(_w, *rfl::get<_i>(_view), new_parent);;
+          s.variant_.visit([&](auto& value) {
+            if constexpr (std::is_same_v<std::remove_cvref_t< decltype(value)>, schema::Type::WithDefault>) {
+              // TODO we have to pass a writer here for other serialization formats
+              // and we would need something like glaze::raw or make default_value_ a template (or std::any, but then we don't know how to serialize
```
autoantwort: @liuzicheng1987 do you have an idea how to handle this?

liuzicheng1987:
@autoantwort, I don't think it's a good idea to be trying to fix two problems at once and fixing #396 is tricky. There is a reason no one really took it up in years.

Also I am very concerned about the proposed the solution of passing a void ptr. 

I didn't know you were going to propose a fix, so I started working on my own: 

https://github.com/getml/reflect-cpp/pull/681

This fixes #679, but still needs some adaptions for some non-JSON formats.

liuzicheng1987:
@autoantwort, if you want to fix #396, I think the best way would be to transform the struct with the default values into an rfl::Generic and then pass that to the to_schema function. That might actually work.

autoantwort:
> @autoantwort, if you want to fix https://github.com/getml/reflect-cpp/issues/396, I think the best way would be to transform the struct with the default values into an rfl::Generic and then pass that to the to_schema function. That might actually work.

Yeah that worked. 

> I didn't know you were going to propose a fix, so I started working on my own: https://github.com/getml/reflect-cpp/pull/681

Ok so you want to continue yours and I than rebase this on yours or you use the idea from here in #681 or in a new PR?

liuzicheng1987:
@autoantwort, I think I can continue on mine to fix #679 and then we can dedicate this PR to fixing #396

liuzicheng1987:
@autoantwort I have just merged my PR which fixes #679, you can merge main into this branch and then continue working on #396.

autoantwort:
Done

liuzicheng1987:
@autoantwort, that's great work. We'll see how the tests do, but I am quite happy.

liuzicheng1987:
@autoantwort, there appear to be minor issues with missing includes (we always want to make sure that every header is self-contained). Could you fix that?

liuzicheng1987:
@autoantwort , merged with a minor edit. Thanks for the contribution!

PR #688, opened 2026-07-02
Avoid clang 21/22 C++26-mode frontend crash in StringLiteral's variadic constructor

aruisdante:
## Problem

Any translation unit that includes reflect-cpp headers fails to compile at `-std=c++26` on clang 21.1+, 22.1.x, and current trunk: the compiler itself crashes (SIGSEGV in `CheckParameterPacksForExpansion`; assertion builds fail with `"declaration not instantiated in this scope"` in `LocalInstantiationScope::findInstantiationOf`) while parsing `using` aliases to `Literal<...>` / `NamedTuple<...>` types — e.g. `parsing/Parser_duration.hpp` and `parsing/Parser_result.hpp` are the first two headers hit. `-std=c++23` and earlier are unaffected.

Root cause is an open clang regression from the CWG2369 reapplication, not a reflect-cpp bug: llvm/llvm-project#198052, reported against reflect-cpp specifically in llvm/llvm-project#205000 (an 11-line minimized reproducer, reduced from `rfl::internal::StringLiteral`, is attached there). The trigger is `StringLiteral`'s variadic constructor constraining itself with a `requires`-fold over `decltype` of its own function parameter pack, alongside the competing `const char (&)[N]` constructor.

## Change

Constrain via template type parameters instead:

```cpp
template <std::same_as<char>... Chars>
constexpr StringLiteral(const Chars... _chars) : arr_{_chars..., '\0'} {}
```

Semantically equivalent — each argument must still be exactly `char` — and it sidesteps the clang bug on every version tested. Since the crash fires at parse time, this one constructor is the only change needed to make the whole library usable at C++26 on affected clangs; it can be reverted freely once fixed clang releases are common.

## Testing

Verified with clang 22.1.7 (libc++, `-fsyntax-only` and full builds): with this change, including `rfl/json.hpp`/`rfl/cbor.hpp` compiles cleanly at `-std=c++20/23/26`, and CBOR + JSON round-trips exercising `Literal`, `Field`, `TaggedUnion`, enums, optionals, and containers pass, including under ASan/UBSan. Please let your CI run the full suite to confirm.

## Disclosure

This fix was authored with an AI assistant (Claude Code) and reviewed/submitted by the account owner.

### DIFF
--- a/include/rfl/internal/StringLiteral.hpp
+++ b/include/rfl/internal/StringLiteral.hpp
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include <array>
+#include <concepts>
 #include <string>
 #include <string_view>
 
@@ -14,9 +15,11 @@ namespace internal {
 /// for the parameters names in the NamedTuples.
 template <size_t N>
 struct StringLiteral {
-  constexpr StringLiteral(const auto... _chars)
-    requires (std::is_same_v<decltype(_chars), const char> && ...)
-    : arr_{_chars..., '\0'} {}
+  /// Constrained via template type parameters rather than a requires-fold
+  /// over decltype of the function parameter pack: the latter crashes
+  /// clang 21/22 in C++26 mode (llvm/llvm-project#198052, #205000).
+  template <std::same_as<char>... Chars>
+  constexpr StringLiteral(const Chars... _chars) : arr_{_chars..., '\0'} {}
 
   constexpr StringLiteral(const std::array<char, N> _arr) : arr_(_arr) {}
 


liuzicheng1987:
@aruisdante your solution fails for macos-latest. Could you take another look? Maybe we need to implement a version-specific solution….

aruisdante:
@liuzicheng1987 that's odd, I was working on MacOS 26.5 when Claud created the patch. When I have a moment I'll look at it and see what's up.

aruisdante:
## The `macos-latest (tests)` failure is unrelated to this PR

Every other matrix cell is green — all Linux (gcc 11–14, llvm 16–18, C++20/23), all Windows/MSVC formats, `macos-15-intel (tests)`, and both `benchmarks` cells. This diff only touches `include/rfl/internal/StringLiteral.hpp`, and the build compiled all of that header's dependent parsers successfully before dying on an unrelated CBOR test file:

```
tests/cbor/test_read_byte_containers.cpp:34
  std::basic_string_view<std::byte> byte_view(...);
error: 'char_traits<std::byte>' is deprecated: char_traits<T> for T not equal to
char, wchar_t, char8_t, char16_t or char32_t is non-standard ... [-Werror,-Wdeprecated-declarations]
```

**Root cause (toolchain, not this PR):** GitHub bumped the `macos-latest` image to macOS 15 + Xcode 26.5, whose libc++ now marks `std::char_traits<T>` deprecated for any `T` outside `{char, wchar_t, char8_t, char16_t, char32_t}`. `std::basic_string_view<std::byte>` instantiates the now-deprecated `std::char_traits<std::byte>`, and tests build with `-Werror`, so it's a hard error. Only this one cell fails because it is the sole intersection of "new libc++" and "compiles the test suite" (`macos-15-intel` runs an older toolchain; the `benchmarks` cells never build the test files). `main` is green only because its last macOS run predates the image bump — it will fail there too on the next run.

**Fix:** opened #691, which reads the CBOR bytes through `std::span<const std::byte>` (idiomatic non-owning byte view, no `char_traits` dependency, satisfies the same `ContiguousByteContainer` concept `rfl::cbor::read` requires). Once that lands, rebasing this branch turns the macOS cell green. Nothing in this PR needs to change.

---

*Disclosure: this investigation was produced with the assistance of an AI coding agent (Claude Code / Claude Opus), reviewed by me before posting.*

liuzicheng1987:
@aruisdante I will merge your other PR, then push main into this one and merge it as well.

liuzicheng1987:
@aruisdante now all of the tests pass. I am merging. Thank you so much for your contribution and sorry that it took a bit longer than usual to merge your PR.

PR #696, opened 2026-07-20
Add operator== for rfl::Generic and rfl::Object

### LINKED ISSUE #693: Equality operator for rfl::Generic and rfl::Object
Hello, we sometimes embed Generic inside of types like:

```cpp
struct Foo {
  std::string x;
  rfl::Generic extras;
};
```

But then, in order to default `operator==` for `Foo`, we need to define our own `operator==` for `rfl::Generic`. The same thing applies to `rfl::Object`. Would it be possible to provide that as part of the library? If this sounds ok I'm happy to submit a PR.

Thanks.

jinchengJL:
Resolves #693 

This adds `operator==` support to `Generic` and `Object`.

Note for `rfl::Object`, I made the comparison _positional_ (same key-value pairs in the same order). This is because `Object` allows for duplicate keys, so order-dependent comparison is more natural. A non-positional comparison would be `O(n^2)` in the worst case, which seems undesirable. Let me know if this is acceptable.

Code was generated with help of Claude. Above note is human-written.

### DIFF
--- a/docs/object.md
+++ b/docs/object.md
@@ -63,3 +63,21 @@ There are three different ways of accessing fields:
 - `.get(...)` returns an `rfl::Result` wrapping the field, or an `rfl::Error` if the field doesn't exist.
 
 Note that it is most efficient, if you access fields in the order that they were placed.
+
+Two objects can be compared for equality using `operator==`. Because `rfl::Object`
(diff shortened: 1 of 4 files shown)


liuzicheng1987:
@jinchengJL Looks great. Thanks for your contribution. I am merging.

PR #707, opened 2026-07-25
Allow validated fields to work alongside defaulted ones

### LINKED ISSUE #700: Reading JSON into a struct with a `DefaultVal` field and a size validator field always raises a size validation error
For example, using `rfl::json::read` with the struct
```c++
struct Person {
  rfl::DefaultVal<std::string> name = "Homer Simpson";
  rfl::Validator<std::vector<std::string>, rfl::Size<rfl::Minimum<1>>> favorite_foods;
};
```
seems to always fail with the exception `Size validation failed: Value expected to be greater than or equal to 1, but got 0.`, even when given a list of `favorite_foods` with one or more items.

A short test case to see the behavior:
```c++
namespace test_default_val_alongside_size_validator {

struct Person {
  rfl::DefaultVal<std::string> name = "Homer Simpson";
  rfl::Validator<std::vector<std::string>, rfl::Size<rfl::Minimum<1>>> favorite_foods;
};

TEST(json, test_default_val_alongside_size_validator) {
  static_assert(rfl::internal::has_default_val_v<Person>,
                "Should have default val");

  const auto should_fail = rfl::json::read<Person>(R"({})");

  EXPECT_EQ(should_fail && true, false)
      << "Should have failed due to missing required field";

  const auto homer =
      rfl::json::read<Person>(R"({"favorite_foods": ["donuts", "pizza"]})").value();

  EXPECT_EQ(homer.name.value(), "Homer Simpson");
  EXPECT_EQ(homer.favorite_foods.value().size(), 2);
  EXPECT_EQ(homer.favorite_foods.value()[0], "donuts");
  EXPECT_EQ(homer.favorite_foods.value()[1], "pizza");
}
}  // namespace test_default_val_alongside_size_validator
```

However, if you set a default value for the field that satisfies the validator, everything seems to work. For example, you can change the `favorite_foods` definition in the above test to
```c++
  rfl::Validator<std::vector<std::string>, rfl::Size<rfl::Minimum<1>>> favorite_foods = {{std::string{"beer"}}};
```
and it should pass.

wbernoudy:
Fixes #700 

Like outlined in my comment in that issue, this works by giving the constructor of `rfl::Validator` an "escape-hatch" using `std::source_location`. If the constructor sees that the caller is specifically the method `rfl::parsing::Parser::read_struct_with_default()`, it will skip validation, allowing the construction of the user's struct with all defaults.

This is definitely a hack and there may be dangers I am unaware of with this, but I thought I would make a PR to see if it was possible (and how bad it was!).

This is also explicitly requiring `std::source_location`; though compiler support for this is required by the library, I see there is some fallback code if that is not available. I am happy to re-work this PR a bit to allow a fallback (or at least not break compilation).

### DIFF
--- a/include/rfl/Validator.hpp
+++ b/include/rfl/Validator.hpp
@@ -11,6 +11,13 @@
 #include "Result.hpp"
 #include "internal/HasValidation.hpp"
 
+namespace rfl::internal::parsing {
+// Forward declare this method used to determine if validation is called from
+// the Parser
+constexpr bool called_from_read_struct_with_default(
+    const std::source_location& loc);
+}  // namespace rfl::internal::parsing
+
 namespace rfl {
 
 /// A wrapper that validates a value against one or more validation rules.
@@ -40,8 +47,12 @@ struct Validator {
   }
 
   /// Default constructor - validates a default-constructed T.
-  /// @throws std::exception if validation of the default value fails
-  Validator() : value_(ValidationType::validate(T()).value()) {}
+  /// @throws std::exception if validation of the default value fails, unless
+  ///   the constructor is being called from
+  ///   `rfl::Parser::read_struct_with_default()`. See the documentation of that
+  ///   method for more details.
+  Validator(const std::source_location& loc = std::source_location::current())
+      : value_(validate_from_default_constructor_(T(), loc)) {}
 
   /// Move constructor.
   /// @param _other The validator to move from
@@ -170,6 +181,16 @@ struct Validator {
   const T& reflection() const { return value_; }
 
  private:
+  T&& validate_from_default_constructor_(T&& value,
+                                         const std::source_location& loc) {
+    // If this was originally called from
+    // `rfl::parsing::Parser::read_struct_with_default()`, then skip validation
+    if (internal::parsing::called_from_read_struct_with_default(loc)) {
+      return std::move(value);
+    }
+    return ValidationType::validate(std::move(value)).value();
+  }
+
   /// The underlying value.
   T value_;
 };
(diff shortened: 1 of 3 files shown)


liuzicheng1987:
@wbernoudy there are multiple build failures in the test pipeline, could you take a look?

Also, what if people use the DefaultIfMissing processor? What if they wrap their validator in DefaultValue? In that case, the validator would no longer guarantee that the condition it sets always holds.

Also, it introduces brittleness to the code - by moving around the code a bit, you could inadvertently invalidate all of the validators.

To be honest, I am also a bit uncomfortable with the solution here...I understand that we use source_location in other places, but we do so in a very controlled manner and it is absolutely essential to the functionality of the library. In this case, it really solves a very minor problem (all you have to do is assign a default value to the field that does fulfill the validator's condition and then there's no problem), so I am unsure whether this is the right way forward and might create more problems in the future than it solves.

wbernoudy:
> Also, what if people use the DefaultIfMissing processor? What if they wrap their validator in DefaultValue? In that case, the validator would no longer guarantee that the condition it sets always holds.

You're right, with the current solution, if you provide a some data to be deserialized that does with that field missing, and the default value for the field does not pass validation, nothing will catch it. I think there might be an approach to solve this: after setting the fields in the parser, we could then "reset" the default ones, which would then cause validation to run on all the fields that have it. But that is pretty messy and probably inefficient as well (and I'm not even sure it would work!).

Agreed on both your other points. I really don't like this solution overall that much either.

Unless you really love the above idea, I'm inclined to close this.

liuzicheng1987:
@wbernoudy I am closing this. But thanks for the initiative anyway. If you have another idea for this problem or another one, an issue and/or PR is always welcome.

PR #720, opened 2026-08-28
Reject externally tagged variants with more than one field

ciprianf-flexion:
Reading an `rfl::Variant<rfl::Field<...>, ...>` from an object with several keys silently keeps the last recognized tag. Worse, an unknown key followed by a valid tag overwrites the "could not match field" error, so `{"triangle": {...}, "circle": {...}}` parses successfully as a circle.

Serde, which this format models (per [variants_and_tagged_unions.md](https://rfl.getml.com/variants_and_tagged_unions/)), represents an externally tagged enum as a single-key map and rejects anything else, see [serde.rs/enum-representations.html](https://serde.rs/enum-representations.html#externally-tagged).

This change makes the reader return an error whenever a second field is encountered, mirroring the existing "expected the object to have exactly one field, but found none" error.

Added tests for both cases in `tests/json/test_field_variant_multiple_fields.cpp`.

### DIFF
--- a/include/rfl/parsing/FieldVariantReader.hpp
+++ b/include/rfl/parsing/FieldVariantReader.hpp
@@ -43,6 +43,12 @@ class FieldVariantReader {
    */
   void read(const std::string_view& _disc_value,
             const InputVarType& _var) const noexcept {
+    if (*field_variant_) {
+      *field_variant_ = error(
+          "Could not parse rfl::Variant: Expected the object to have "
+          "exactly one field, but found more than one.");
+      return;
+    }
     try_matching_fields(
         _disc_value, _var,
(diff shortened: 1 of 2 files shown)


ciprianf-flexion:
@liuzicheng1987 I cannot add you explicitly as a reviewer, but would you mind taking a look? Thanks!

liuzicheng1987:
@ciprianf-flexion you don't have to explicitly add me as a reviewer, I will take a look.

liuzicheng1987:
@ciprianf-flexion good catch. I never really thought of that problem.

liuzicheng1987:
@ciprianf-flexion thanks for the contribution. I am merging.

Issue #689, opened 2026-07-03
Enum-Flag Default Values


PapeCoding:
Currently I try to use reflect-cpp to serialize a struct to json and back.

```cpp
enum class PlatformFlag
{
	NONE = 0,
	LINUX = 1 << 0,
	WINDOWS = 1 << 1,
	MAC = 1 << 2,
	ALL = LINUX | WINDOWS | MAC
};
constexpr PlatformFlag operator&(PlatformFlag const a, PlatformFlag const b);
constexpr PlatformFlag operator|(PlatformFlag const a, PlatformFlag const b);
inline PlatformFlag& operator|=(PlatformFlag& a, PlatformFlag const b);

[...]

struct myStruct
{
	PlatformFlag supportedPlatforms = PlatformFlag::NONE;
};
```

Serializing to json without setting a value yields something like:
```json
{
    "supportedPlatforms": ""
}
```
and reading it back then errors out with: `Invalid enum value: ''. Must be one of [NONE, LINUX, WINDOWS, MAC].`.

Am I missing something? Ideally it would either serialize to "NONE" or allow the empty string here.

liuzicheng1987:
@PapeCoding what version are you using? I am pretty sure we have seen this before and fixed it already.

liuzicheng1987:
@PapeCoding I cannot reproduce the issue on the newest version of reflect-cpp. Here is the test I wrote:

```cpp
enum class PlatformFlag {
  NONE = 0,
  LINUX = 1 << 0,
  WINDOWS = 1 << 1,
  MAC = 1 << 2,
  ALL = LINUX | WINDOWS | MAC
};

constexpr PlatformFlag operator&(PlatformFlag const a, PlatformFlag const b) {
  return static_cast<PlatformFlag>(static_cast<uint64_t>(a) &
                                   static_cast<uint64_t>(b));
}

constexpr PlatformFlag operator|(PlatformFlag const a, PlatformFlag const b) {
  return static_cast<PlatformFlag>(static_cast<uint64_t>(a) |
                                   static_cast<uint64_t>(b));
}

inline PlatformFlag& operator|=(PlatformFlag& a, PlatformFlag const b) {
  a = a | b;
  return a;
}

struct myStruct {
  PlatformFlag supportedPlatforms = PlatformFlag::NONE;
};

TEST(json, test_enum9) {
  myStruct t{.supportedPlatforms = PlatformFlag::NONE};

[...]

PR #721, opened 2026-08-29
Output zero flag name if available

### LINKED ISSUE #640: Problem (de-)serializing an enum with value 0 if the | operator is overloaded
Hello,
I am unfortunately running into an issue with the newest version of the reflect-cpp lib.

Assume I have something like this:
```
enum class TestEnum : uint64_t
{
	None = 0,
	Hello = 1 << 0,
	World = 1 << 1,
	HelloWorld = Hello | World
};
struct SomeClass
{
	TestEnum e;
	TestEnum f;
	TestEnum g;
};

void Foo()
{
	SomeClass t{.e = TestEnum::None, .f = TestEnum::Hello, .g = TestEnum::HelloWorld};
	const auto jsonString = rfl::json::write(t);
	const auto t2 = rfl::json::read<SomeClass>(jsonString).value();
}
```

^Works fine. The object will be serialized to `{"e":"None","f":"Hello","g":"HelloWorld"}` and back as expected.

The issue arises once I overload the `|` operator:
```
static constexpr TestEnum operator|(TestEnum a, TestEnum b) noexcept { 
	return static_cast<TestEnum>(static_cast<uint64_t>(a) | static_cast<uint64_t>(b)); 
}
```
This changes the serialization to `{"e":"","f":"Hello","g":"Hello|World"}`. The `"Hello|World"` part is expected, as per your documentation.

The issue is that what was previously `"None"` was now serialized to the empty string `""`, eventually failing the de-serialization step with the error "Failed to parse field 'e': Invalid enum value: ''. Must be one of [None, Hello, World].".

I am unsure what the expected behavior should be. I can see all variants `"None"`, `""` and `"0"` make sense. Imo the bigger problem is the fact that the immediate de-serialization fails.

I would be very thankful for your help!
Best, Florian

ZXShady:
This makes it so enums output the name of the zero flag if it exists fixes #640 and #689

### DIFF
--- a/include/rfl/enums.hpp
+++ b/include/rfl/enums.hpp
@@ -72,7 +72,7 @@ std::string enum_to_string(const EnumType _enum) {
       val >>= 1;
     }
     if (flags.empty()) {
-      return "0";
+      return internal::enums::to_string(static_cast<EnumType>(0));
     }
     return internal::strings::join("|", flags);
   } else {
--- a/tests/json/test_enum8.cpp
+++ b/tests/json/test_enum8.cpp
@@ -32,7 +32,7 @@ TEST(json, test_enum8) {
   SomeClass t{
       .e = TestEnum::None, .f = TestEnum::Hello, .g = TestEnum::HelloWorld};
 
-  write_and_read(t, R"({"e":"0","f":"Hello","g":"Hello|World"})");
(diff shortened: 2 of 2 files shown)


liuzicheng1987:
@ZXShady thanks for the contribution!

PR #725, opened 2026-09-18
fix `Generic.cpp` build error due to missing includes

gg-sr:
I'm not sure why it works for others, but not for me. Maybe different compilation flags, since I'm wrapping the library in Bazel and not using the provided CMakeLists.

### DIFF
--- a/src/rfl/Generic.cpp
+++ b/src/rfl/Generic.cpp
@@ -26,6 +26,11 @@ SOFTWARE.
 
 #include "rfl/Generic.hpp"
 
+#include <cstdint>
+#include <limits>
+#include <optional>
+#include <variant>
+
 namespace rfl {
 
 Generic::Generic() : value_(std::nullopt) {}


liuzicheng1987:
@gg-sr that is weird, because we have pipeline checks just for that. But thanks for the contribution!

gg-sr:
Here is the error I was getting:

```
external/+http_archive+reflect-cpp/src/rfl/Generic.cpp:112:42: error: implicit instantiation of undefined template 'std::numeric_limits<int>'
  112 |     if (*ptr < static_cast<int64_t>(std::numeric_limits<int>::min()) ||
      |                                          ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/max_size_type.h:53:10: note: template is declared here
   53 |   struct numeric_limits;
      |          ^
external/+http_archive+reflect-cpp/src/rfl/Generic.cpp:113:42: error: implicit instantiation of undefined template 'std::numeric_limits<int>'
  113 |         *ptr > static_cast<int64_t>(std::numeric_limits<int>::max())) {
      |                                          ^
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/max_size_type.h:53:10: note: template is declared here
   53 |   struct numeric_limits;
      |          ^
```

Adding `#include <limits>` fixes it. I added the other includes for completeness.

I assume you (or one of your dependencies) sets a flag that implicitly includes needed `std` headers, or _our_ flags forbid such implicit includes, or our toolchains have different transitive includes.

